PR #5, opened 2023-01-19
Add cloudformation bucket stacks

philerooski:


thomasyu888:
Thanks @philerooski this is a good example of a small story ticket that could have been assigned to Rixing. let's think about that moving forward with our work to perhaps chunk things down even further.

PR #6, opened 2023-01-19
[ETL-283] S3 to JSON job

philerooski:
* New stacks for S3 to JSON Glue job, The IAM role for the job, and the Glue workflow which will serve as our primary (and likely only) workflow.
* New stack for pilot data S3 bucket.
* S3 to JSON job implemented

### DIFF
--- a/src/glue/jobs/s3_to_json.py
+++ b/src/glue/jobs/s3_to_json.py
@@ -1,36 +1,34 @@
+"""
+S3 to JSON job
+
+Bucket/sort JSON from a zip archive provided by Care Evolution into
+JSON datasets in S3. A JSON dataset is specific to a key prefix
+(for example: json/dataset=EnrolledParticipants/) and only contains
+files which share a similar schema.
+"""
 import datetime
 import io
 import os
 import json
 import logging
+import sys
 import zipfile
 import boto3
-#from awsglue.utils import getResolvedOptions
+from awsglue.utils import getResolvedOptions
 
 def write_file_to_json_dataset(z, json_path, dataset_identifier,
         metadata, workflow_run_properties):
     """
     Write a JSON from a zipfile to a JSON dataset.
 
-    Additional fields are inserted at the top-level (if the JSON is an object) or
-    at the top-level of each object (if the JSON is a list of objects) before
-    the JSON is written as NDJSON (e.g., since this is a single JSON document,
-    on a single line) to a JSON dataset in S3. Partition fields are inserted into
-    every JSON, regardless of its schema identifier. Every field from the S3
-    object metadata is inserted into the JSON with schema identifier
-    'ArchiveMetadata'.
-
-    Partition fields are as follows:
-        * assessmentid
-        * year (as derived from the `uploadedon` field of the S3 object metadata)
-        * month ("")
-        * day ("")
+    Metadata fields derived from the file basename are inserted as top-level fields
+    and the NDJSON is written to a JSON dataset in S3.
 
     Args:
-        z (zipfile.Zipfile): The zip archive as provided by Bridge.
-        json_path (str): A path relative to the root of `z` to a JSON file.
-        dataset_identifier (str): A BridgeDownstream dataset identifier.
-        s3_obj_metadata (dict): A dictionary of S3 object metadata.
+        z (zipfile.Zipfile): The zip archive as provided by the data provider.
+        json_path (str): A JSON path relative to the root of `z`.
+        dataset_identifier (str): The data type of `json_path`.
+        metadata (dict): Metadata derived from the file basename.
         workflow_run_properties (dict): The workflow arguments
 
     Returns:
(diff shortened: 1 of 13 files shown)


src/glue/jobs/s3_to_json.py
```diff
@@ -0,0 +1,209 @@
+"""
+S3 to JSON job
+
+Bucket/sort JSON from a zip archive provided by Care Evolution into
+JSON datasets in S3. A JSON dataset is specific to a key prefix
+(for example: json/dataset=EnrolledParticipants/) and only contains
+files which share a similar schema.
+"""
+import datetime
```
thomasyu888, line 9: Nit: Please follow the best practices for sorting imports: https://peps.python.org/pep-0008/#imports.  There is vscode module that you can use that automatically does this as well.

https://marketplace.visualstudio.com/items?itemName=ms-python.isort.

By default you can also use the `black` vscode module to auto lint and structure your code.

thomasyu888, line 9: Nit: Please follow the best practices for sorting imports: https://peps.python.org/pep-0008/#imports. There is vscode module that you can use that automatically does this as well.

https://marketplace.visualstudio.com/items?itemName=ms-python.isort.

By default you can also use the black vscode module to auto lint and structure your code.

src/glue/jobs/s3_to_json.py
```diff
@@ -0,0 +1,209 @@
...
+import io
+import os
+import json
+import logging
+import sys
+import zipfile
+import boto3
+from awsglue.utils import getResolvedOptions
+
+def write_file_to_json_dataset(z, json_path, dataset_identifier,
```
thomasyu888, line 19: Nit: If you use type hints, vscode has the ability to actually highlight where it may think you're using the wrong types.

```
def write_file_to_json_dataset(z: zipfile.Zipfile, json_path: str, dataset_identifier: str...
```

src/glue/jobs/s3_to_json.py
```diff
@@ -0,0 +1,209 @@
...
+    with open(output_path, "rb") as f_in:
+        response = s3_client.put_object(
+                Body = f_in,
+                Bucket = workflow_run_properties["json_bucket"],
+                Key = s3_output_key,
+                Metadata = s3_metadata)
+        logger.debug(f"S3 Put object response: {json.dumps(response)}")
+    return output_path
+
+def get_metadata(basename):
```
thomasyu888, line 92:
```suggestion
def get_metadata(basename: str) -> dict:
```

thomasyu888: When you add the type hints, vscode has a module that can auto generate the docstrings actually.  I would look into that.

src/glue/jobs/s3_to_json.py
```diff
@@ -0,0 +1,209 @@
...
+        workflow_run_properties (dict): The workflow arguments
+
+    Returns:
+        output_path (str) The local path the file was written to.
+    """
+    s3_client = boto3.client("s3")
+    logger = logging.getLogger(__name__)
+    os.makedirs(dataset_identifier, exist_ok=True)
+    s3_metadata = metadata.copy()
+    if s3_metadata["start_date"] is None:
```
thomasyu888, line 41: Does the `start_date` key always exist?

src/glue/jobs/s3_to_json.py
```diff
@@ -0,0 +1,209 @@
...
+    if "-" in basename_components[-1]:
+        start_date, end_date = basename_components[-1].split("-")
+        metadata["start_date"] = \
+                datetime.datetime.strptime(start_date, "%Y%m%d")
+        metadata["end_date"] = \
+                datetime.datetime.strptime(end_date, "%Y%m%d")
+    else:
+        metadata["start_date"] = None
+        metadata["end_date"] = \
+                datetime.datetime.strptime(basename_components[-1], "%Y%m%d")
```
thomasyu888, line 122: Could you provide an example of what a basename looks like?

src/glue/jobs/s3_to_json.py
```diff
@@ -0,0 +1,209 @@
...
+        metadata["subtype"] = basename_components[1]
+    if (
+            metadata["type"] == "HealthKitV2Samples"
+            and basename_components[-2] == "Deleted"
+        ):
+        metadata["type"] = "HealthKitV2Samples_Deleted"
+    logger.debug(f"metadata = {metadata}")
+    return metadata
+
+def process_record(s3_obj, workflow_run_properties):
```
thomasyu888, line 133:
```suggestion
def process_record(s3_obj: dict, workflow_run_properties: dict):
```

src/glue/jobs/s3_to_json.py
```diff
@@ -0,0 +1,209 @@
...
+                    dataset_identifier=dataset_identifier,
+                    metadata=metadata,
+                    workflow_run_properties=workflow_run_properties)
+
+def main():
+    # Configure logger
+    logging.basicConfig()
+    logger = logging.getLogger(__name__)
+    logger.setLevel(logging.DEBUG)
+
```
thomasyu888, line 174: Just out of curiosity - Is there a reason why you set the logger here and now on top of the module?

templates/glue-job-S3ToJsonS3.yaml
```diff
@@ -0,0 +1,85 @@
+AWSTemplateFormatVersion: '2010-09-09'
```
thomasyu888, line 1: Out of curiousity: Is this the same template as whats in BridgeDownstream?

thomasyu888 (changes requested):
Great PR!  Exciting to see this.

@rxu17, could you also give this a review?

src/glue/jobs/s3_to_json.py
philerooski: I don't use vscode. But if you know any vim plugins that do the same thing feel free to share.

src/glue/jobs/s3_to_json.py
philerooski: No, not for data types which don't include a start date in their file name: https://sagebionetworks.jira.com/wiki/spaces/IBC/pages/2789376001/Notes#Types-of-data-included-in-archive

src/glue/jobs/s3_to_json.py
philerooski: See the page I linked you to above: https://sagebionetworks.jira.com/wiki/spaces/IBC/pages/2789376001/Notes#Types-of-data-included-in-archive

For example, `HealthKitV2Samples_ActiveEnergyBurned_20220111-20230103.json`

src/glue/jobs/s3_to_json.py
philerooski: This logger is scoped to the `main` function. So each function needs to get the existing logger (which is only configured once here) if they want to use logger methods.

templates/glue-job-S3ToJsonS3.yaml
philerooski: Yes.

src/glue/jobs/s3_to_json.py
thomasyu888: Highly recommend the switch to vscode, but here are some vim plugins I found by doing a quick google search.
* https://vimawesome.com/plugin/vim-isort
* https://vimawesome.com/plugin/black

rxu17 (commented):
Lgtm!

src/glue/jobs/s3_to_json.py
thomasyu888: Thanks - I'll give that page a read!  Thanks!

templates/glue-job-role.yaml
```diff
@@ -0,0 +1,130 @@
...
+          Version: '2012-10-17'
+          Statement:
+          - Effect: Allow
+            Action:
+            - s3:Get*
+            - s3:List*
+            Resource:
+            - !Sub arn:aws:s3:::${S3SourceBucketName}
+            - !Sub arn:aws:s3:::${S3SourceBucketName}/*
+      - PolicyName: EC2
```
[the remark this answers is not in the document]
philerooski: The Glue job here is operating in the same way as the glue job in Bridge Downstream, so I wouldn't expect any of these permissions to change. That said, I don't know if it's using _every_ permission listed here because I took the EC2 permissions from the Glue service policy provided by AWS.

philerooski (commented):
Include tests.

thomasyu888 (approved):
Looks good @philerooski !  Thanks!  Let's try to address the Nit comments if possible.  Since you use `vim`, I suggest adding a ticket to implement `pre-commit`.  We can then add `black` and `isort` which will automatically run upon commits.

philerooski:
Skipping tests for now because of upcoming deadline to produce pilot dataset. Tracking [here](https://sagebionetworks.jira.com/browse/ETL-325).

PR #10, opened 2023-01-27
[ETL-294] create glue table stack

rxu17:
**Purpose:** Creates a glue tables' stack along with relevant configs and table schemas that dynamically create glue tables for the recover datasets for each data type

- Glue tables stack
- Glue table schemas


templates/glue-tables.j2
```diff
@@ -0,0 +1,60 @@
...
+  S3IntermediateBucketName:
+    Type: String
+    Description: Name of the S3 intermediate (JSON) bucket
+
+Resources:
+  # combine all configuration data into a "datasets" collection
+  {% set datasets = [] %}
+  {% for v in sceptre_user_data.dataset_schemas.tables.keys() %}
+  {% set dataset = {} %}
+  {% do dataset.update({'dataset_name': v}) %}
```
thomasyu888, line 23: I'm looking at the template in BridgeDownstream, can you explain why these [lines](https://github.com/Sage-Bionetworks/BridgeDownstream/blob/8994ffde9240175d36488f5ccb8cba8e5667390e/templates/study-pipeline-infra.j2#L102-L106) aren't needed? 

Is it because there are no crawlers?

thomasyu888 (approved):
🔥   Looks great @rxu17 .  I'll let @philerooski take another pass at this.

templates/glue-tables.j2
rxu17: Yes currently, we don't have any plans for glue crawlers or partitions in the glue tables.

templates/glue-tables.j2
```diff
@@ -0,0 +1,60 @@
...
+
+# buckets
+  S3IntermediateBucketName:
+    Type: String
+    Description: Name of the S3 intermediate (JSON) bucket
+
+Resources:
+  # combine all configuration data into a "datasets" collection
+  {% set datasets = [] %}
+  {% for v in sceptre_user_data.dataset_schemas.tables.keys() %}
```
philerooski, line 21: Where does this value come from?

src/glue/resources/table_columns.yaml
```diff
@@ -0,0 +1,461 @@
...
+      - Name: Value
+        Type: string
+      - Name: Units
+        Type: string
+      - Name: Source
+        Type: struct<Identifier:string,Name:string,Version:string,OperatingSystemVersion:string,ProductType:string>
+      - Name: Device
+        Type: struct<Name:string,Model:string,Manufacturer:string,HardwareVersion:string,SoftwareVersion:string,FirmwareVersion:string,LocalIdentifier:string,FDAIdentifier:string>
+      - Name: Metadata
+        Type: struct<HKMetadataKeySyncVersion:string,HKVO2MaxTestType:string,HKMetadataKeySyncIdentifier:string>
```
philerooski, line 77: For these arbitrary JSON fields, reference the tables in `test-database` that I created using the crawlers. You'll find a few more fields there. For example, this is the struct for the Metadata field:

```
{
  "metadata": {
    "HKWasUserEntered": "string",
    "HKAlgorithmVersion": "string",
    "HKMetadataKeyHeartRateMotionContext": "string",
    "HKDateOfEarliestDataUsedForEstimate": "string",
    "HKMetadataKeyAppleDeviceCalibrated": "string",
    "HKTimeZone": "string",
    "HKMetadataKeySyncVersion": "string",
    "HKVO2MaxTestType": "string",
    "HKMetadataKeySyncIdentifier": "string",
    "HKMetadataKeyDevicePlacementSide": "string"
  }
}
```

src/glue/resources/table_columns.yaml
```diff
@@ -0,0 +1,461 @@
+tables:
+  EnrolledParticipants:
```
philerooski, line 2: Every table needs at least one of `export_start_date` or `export_end_date` fields. If only a single date is included in the file name (as it is for the EnrolledParticipants data type), then only `export_end_date` is needed. Otherwise both fields are needed.

templates/glue-tables.j2
```diff
@@ -0,0 +1,60 @@
...
+  {% do dataset.update({'stackname_prefix': '{}'.format(v.replace('_',''))}) %}
+  {% do datasets.append(dataset) %}
+  {% endfor %}
+
+  GlueDatabase:
+    Type: AWS::Glue::Database
+    Properties:
+      CatalogId: !Ref AWS::AccountId
+      DatabaseInput:
+        Description: !Sub 'Recover database'
```
philerooski, line 36: Include a reference to the namespace in the Description

templates/glue-tables.j2
```diff
@@ -0,0 +1,60 @@
...
+        Retention: 0
+        StorageDescriptor:
+          Columns: {{ dataset.columns }}
+          Compressed: false
+          InputFormat: org.apache.hadoop.mapred.TextInputFormat
+          Location: !Sub s3://${S3IntermediateBucketName}/${Namespace}/
+          OutputFormat: org.apache.hadoop.hive.ql.io.HiveIgnoreKeyTextOutputFormat
+          StoredAsSubDirectories: false
+        TableType: EXTERNAL_TABLE
+    {% endfor %}
```
philerooski, line 60: I think this is indented once too much.

philerooski (changes requested):

templates/glue-tables.j2
rxu17: The sceptre_user_data.dataset_schemas.tables.keys? It comes from here: [config/develop/namespaced/glue-tables.yaml](https://github.com/Sage-Bionetworks/recover/blob/etl-294-create-glue-table-stack/config/develop/namespaced/glue-tables.yaml#L12-L14)

src/glue/resources/table_columns.yaml
rxu17: Got it! I think I got confused between seeing if the datatypes themselves have a start date or end date variable in their schemas

templates/glue-tables.j2
rxu17: I copied this from BridgeDownstream and it had it as well. Will make a note to correct it there for syntax purposes. It doesn't appear to affect the code's working ability at least here.

templates/glue-tables.j2
philerooski: But is `dataset_schemas` defined in the `sceptre_user_data`?

rxu17: Yes, it's defined as `src/glue/resources/table_columns.yaml`

philerooski: Ohh, I didn't realize this was set in the tables stack config.

philerooski:
I just noticed that at least one of the table names don't match the data types. For example, what in `table_columns.yaml` is called `HealthKitV2Electrocardiogram_Samples` should be `HealthKitV2Electrocardiogram`. Can you match the data types as they appear in the `recover-dev-intermediate-data` bucket?

philerooski (approved):

PR #14, opened 2023-02-09
[ETL-340] Set up sts access

rxu17:
**Purpose:** Set up parquet datasets for STS access
- Bucket Policy of the S3 bucket template modified so that it writes an owner.txt to the parquet dataset location.
- Script for setting up STS and the storage location to be the S3 location of the parquet datasets

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -8,6 +8,8 @@ python_version = "3.10"
 
 [dev-packages]
 pytest = "*"
+pyarrow = "*"
 pre-commit = "*"
 sceptre = ">=3.2.0"
 sceptre-sam-handler = "*"
+synapseclient = "==2.7.*"
--- a/Pipfile
+++ b/Pipfile
@@ -8,8 +8,8 @@ python_version = "3.10"
 
 [dev-packages]
 pytest = "*"
-pyarrow = "*"
+pyarrow = "~=11.0"
 pre-commit = "*"
 sceptre = ">=3.2.0"
 sceptre-sam-handler = "*"
-synapseclient = "==2.7.*"
+synapseclient = "~=2.7"
--- a/src/scripts/setup_external_storage/setup_external_storage.py
+++ b/src/scripts/setup_external_storage/setup_external_storage.py
@@ -18,15 +18,17 @@ def read_args():
     )
     parser.add_argument(
         "--synapse-folder-name",
+        default="parquet",
         help=(
-            "Name of the Synapse folder to use with the external " "storage location."
+            "Name of the Synapse folder to use with the external "
+            "storage location. Default is parquet"
         ),
     )
     parser.add_argument("--s3-bucket", help="S3 bucket name")
     parser.add_argument(
         "--s3-key",
-        default=None,
-        help="S3 key to serve as the root. `/` (root) by default.",
+        default="recover/parquet",
+        help="S3 key to serve as the root. Default is recover/parquet.",
     )
     parser.add_argument(
         "--sts-enabled",
@@ -40,18 +42,18 @@ def read_args():
 def main():
     args = read_args()
     syn = synapseclient.login()
-    storage_location = syn.create_s3_storage_location(
+    synapse_folder, storage_location, synapse_project = syn.create_s3_storage_location(
         parent=args.synapse_parent,
         folder_name=args.synapse_folder_name,
         bucket_name=args.s3_bucket,
         base_key=args.s3_key,
         sts_enabled=args.sts_enabled,
     )
+
     storage_location_info = {
-        k: v
-        for k, v in zip(
-            ["synapse_folder", "storage_location", "synapse_project"], storage_location
-        )
+        "synapse_folder": synapse_folder,
+        "storage_location": storage_location,
+        "synapse_project": synapse_project,
     }
     print(storage_location_info)
     return storage_location_info
--- a/src/scripts/setup_external_storage/setup_external_storage.py
+++ b/src/scripts/setup_external_storage/setup_external_storage.py
@@ -1,0 +1,61 @@
+"""
+Create an STS-enabled folder on Synapse over an S3 location.
+"""
+
+import os
+import json
+import argparse
+
+import synapseclient
+
+
+def read_args():
+    parser = argparse.ArgumentParser(
+        description="Create an STS-enabled folder on Synapse over an S3 location."
+    )
+    parser.add_argument(
+        "--synapse-parent", help="Synapse ID of the parent folder/project"
+    )
+    parser.add_argument(
+        "--synapse-folder-name",
+        help=(
+            "Name of the Synapse folder to use with the external " "storage location."
+        ),
+    )
+    parser.add_argument("--s3-bucket", help="S3 bucket name")
+    parser.add_argument(
+        "--s3-key",
+        default=None,
+        help="S3 key to serve as the root. `/` (root) by default.",
+    )
+    parser.add_argument(
+        "--sts-enabled",
+        action="store_true",
+        help="Whether this storage location should be STS enabled",
+    )
+    args = parser.parse_args()
+    return args
+
+
+def main():
+    args = read_args()
+    syn = synapseclient.login()
+    storage_location = syn.create_s3_storage_location(
+        parent=args.synapse_parent,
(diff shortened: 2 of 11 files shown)


Pipfile
```diff
@@ -8,6 +8,8 @@ python_version = "3.10"
 
 [dev-packages]
 pytest = "*"
+pyarrow = "*"
 pre-commit = "*"
 sceptre = ">=3.2.0"
 sceptre-sam-handler = "*"
+synapseclient = "==2.7.*"
```
thomasyu888, line 15: Nit: You may want to lock down the dependencies here. 

Instead of doing `==2.7.*` and can do `~=2.7`.  See this [slackoverflow](https://stackoverflow.com/questions/39590187/in-requirements-txt-what-does-tilde-equals-mean)

src/scripts/setup_external_storage/setup_external_storage.py
```diff
@@ -0,0 +1,61 @@
...
+        bucket_name=args.s3_bucket,
+        base_key=args.s3_key,
+        sts_enabled=args.sts_enabled,
+    )
+    storage_location_info = {
+        k: v
+        for k, v in zip(
+            ["synapse_folder", "storage_location", "synapse_project"], storage_location
+        )
+    }
```
thomasyu888, line 55: I like the neat trick here that you do to parse through the tuple `create_s3_storage_location` returns.  The reason I like my suggested correction is it a quicker way to knowing that `create_s3_storage_location` returns a tuple. Thoughts? 

```suggestion
    synapse_folder, storage_location, synapse_project = syn.create_s3_storage_location(
        parent=args.synapse_parent,
        folder_name=args.synapse_folder_name,
        bucket_name=args.s3_bucket,
        base_key=args.s3_key,
        sts_enabled=args.sts_enabled,
    )
    storage_location_info = {
        "synapse_folder": synapse_folder,
        "storage_location": storage_location,
        "synapse_project": synapse_project
    }
```

src/scripts/setup_external_storage/setup_external_storage.py
```diff
@@ -0,0 +1,61 @@
...
+        help="Whether this storage location should be STS enabled",
+    )
+    args = parser.parse_args()
+    return args
+
+
+def main():
+    args = read_args()
+    syn = synapseclient.login()
+    storage_location = syn.create_s3_storage_location(
```
thomasyu888, line 43: Did you copy this script from BridgeDownstream?  I need to add it to my TODOs to create an EPIC for creating a RECOVER / BridgeDownstream utility package.  

1. should this function be returned as a dict in the python client?
2. Will having this command as a cli as part of the client remove the need for this script?

templates/s3-bucket.yaml
```diff
@@ -79,6 +88,21 @@ Resources:
               Resource: !Sub ${Bucket.Arn}/*
             - !Ref AWS::NoValue
 
+  SynapseOwnerFile:
+    Type: AWS::S3::Object
+    Condition: ConnectToSynapse
```
thomasyu888, line 93: You removed the condition up top.  You will have to change this.

tests/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM amazon/aws-glue-libs:glue_libs_3.0.0_image_01
+FROM amazon/aws-glue-libs:glue_libs_4.0.0_image_01
 
-RUN pip3 install pytest-datadir
+RUN pip3 install synapseclient==2.7.* pyarrow==11.* pytest-datadir
```
thomasyu888, line 3:
```suggestion
RUN pip3 install synapseclient~=2.7 pyarrow~=11 pytest-datadir
```

thomasyu888 (changes requested):
🛸 Looks great! Some comments/questions

templates/s3-bucket.yaml
rxu17: Hmm, do these lines not cover it: [L32-L33](https://github.com/Sage-Bionetworks/recover/blob/3c134df081085a0e58a0e3a292ac8a97462ae163/templates/s3-bucket.yaml#L32-L33)?

src/scripts/setup_external_storage/setup_external_storage.py
```diff
@@ -0,0 +1,61 @@
...
+
+def read_args():
+    parser = argparse.ArgumentParser(
+        description="Create an STS-enabled folder on Synapse over an S3 location."
+    )
+    parser.add_argument(
+        "--synapse-parent", help="Synapse ID of the parent folder/project"
+    )
+    parser.add_argument(
+        "--synapse-folder-name",
```
philerooski, line 20: Can we give this a default name? "parquet" is fine.

src/scripts/setup_external_storage/setup_external_storage.py
```diff
@@ -0,0 +1,61 @@
...
+        help=(
+            "Name of the Synapse folder to use with the external " "storage location."
+        ),
+    )
+    parser.add_argument("--s3-bucket", help="S3 bucket name")
+    parser.add_argument(
+        "--s3-key",
+        default=None,
+        help="S3 key to serve as the root. `/` (root) by default.",
```
philerooski, line 29: We know where the production data will be in S3, so let's set a more useful default than the bucket root.

philerooski (approved):

templates/s3-bucket.yaml
thomasyu888: Ah my mistake! Thanks!

src/scripts/setup_external_storage/setup_external_storage.py
rxu17: Is the preference to just get each object of the tuple `storage_location` by index then like this? 
```
storage_location_info = {
        "synapse_folder": storage_location[0],
        "storage_location": storage_location[1],
        "synapse_project": storage_location[2]
    }
```

src/scripts/setup_external_storage/setup_external_storage.py
rxu17: Yes this is copied from here: [src/scripts/setup_external_storage/setup_external_storage.py](https://github.com/Sage-Bionetworks/BridgeDownstream/blob/main/src/scripts/setup_external_storage/setup_external_storage.py) except we don't copy any bucket policy (done in S3 bucket template), create any owner.txt (done in S3 bucket template) or cors rules anymore

src/scripts/setup_external_storage/setup_external_storage.py
rxu17: Yes I agree, the code seems more intuitive as well.

src/scripts/setup_external_storage/setup_external_storage.py
rxu17: 1. I'm looking at the current use case for this function `create_s3_storage_location` and it's not really being used elsewhere other than BridgeDownstream, here (for just this one script) and the python synapse client itself. I don't think it needs to be returned as a dict except for the benefit of just guaranteeing that should the order of the items in the dict change, it's fine since we are pulling them out by the key?
2. What would that look like?

thomasyu888 (approved):
✔️ LGTM!

src/scripts/setup_external_storage/setup_external_storage.py
thomasyu888: 1. I think thats fine - I don't think the order of the keys will change because tests will fail if the order changed on the client.
2. When you install the synapsePythonClient, a command line client is installed as well.  So... something like this in the terminal

    ```
     synapse create-s3-storage-location ...
     ```

rxu17: >     2. When you install the synapsePythonClient, a command line client is installed as well.  So... something like this in the terminal
>        ```
>         synapse create-s3-storage-location ...
>        ```
Oh I see. Does the cli command `synapse login` allow for logging in with an `authToken` instead of cached credentials? But, yes I can see that it would eliminate the use for this script if we didn't plan to expand this script to anything else other than this.

thomasyu888: Got it. I will add this to the backlog of tickets to potentially create and track the usecase.

PR #15, opened 2023-02-15
[ETL-326] block public access in bucket

rxu17:
**Purpose:** Add to cloudformation so bucket creation would involve blocking public access to bucket. 
This includes blocking:
- ACLs in general that tries to grant public access to bucket
- bucket and access point policies that tries to grant public and cross-account access to bucket


thomasyu888 (approved):
🚀 LGTM!

philerooski (approved):

PR #16, opened 2023-02-24
[ETL-341] s3 to json tests

rxu17:
**Purpose**: This contains tests written for the [src/glue/jobs/s3_to_json.py](https://github.com/Sage-Bionetworks/recover/blob/main/src/glue/jobs/s3_to_json.py) script.

The tests focus on these main functions that constitute the heart of the s3_to_json.py code. Some of the tests' skeletons were borrowed from BridgeDownstream:
- get_metadata
- write_file_to_json_dataset

Note that since schema validation wasn't implemented yet for s3 to json, that wasn't implemented into the tests as well

PR #17, opened 2023-02-24
[ETL-328] Encode EnrolledParticipants JSON string field as object

philerooski:
* Encode `.[“CustomFields”][“Symptoms”]` from EnrolledParticipants data type as an object.
* Small fix to S3 bucket template that caused stacks to fail buckets are not granting access to Synapse.


src/glue/jobs/s3_to_json.py
```diff
@@ -60,6 +61,24 @@ def write_file_to_json_dataset(
                 # Load JSON string as dict
                 j["Value"] = json.loads(j["Value"])
             data.append(j)
+            if dataset_identifier == "EnrolledParticipants":
+                if "Symptoms" in j["CustomFields"]:
+                    if len(j["CustomFields"]["Symptoms"]) > 0:
```
thomasyu888, line 66: Nit: This can be rolled up into one if statement:

```
if "Symptoms" in j["CustomFields"] and len(j["CustomFields"]["Symptoms"]) > 0:
```

thomasyu888 (approved):
👍🏼 LGTM! Just a minor comment.

src/glue/jobs/s3_to_json.py
philerooski: It could be, but then I would need to add another `if "Symptoms" in j["CustomFields"]` statement on line 80 and that would be redundant.

thomasyu888: Ah, that makes sense - I should've been more thorough.

philerooski:
Blocked by #16

PR #18, opened 2023-03-07
Write type of HealthKitV2Statistics to JSON

philerooski:
I noticed that the HealthKitV2Statistics data type has a subtype included in its file name. I'm handling this the same way as we already handle the subtype in HealthKitV2Samples.


thomasyu888 (approved):
🚀 LGTM!

PR #20, opened 2023-03-08
[ETL-361] Fix lambda function to use env var due to failing test

rxu17:
**Purpose:** Originally the `lambda function` code hard coded in the `recover dev bucket` and `glue workflow` names and resulted in the tests having to test that those names exist which caused the tests to fail if you switched to the production account or deleted the workflow/changed the name. 

Now the lambda function is more robust as it takes the workflow and bucket names from the cloudformation templates as environment variables and the lambda stack also requires for the glue workflow to exist as a dependency so the tests can be removed.


thomasyu888 (approved):
🔥 nice - this makes sense to me.

philerooski (approved):

PR #21, opened 2023-03-10
[ETL-336] Upload artifacts

rxu17:
**Purpose:** This PR allows us to upload artifacts to the prod and dev S3 cloudformation buckets as part of the upload_and_deploy CI/CD workflow in github using the new OIDC integration.


thomasyu888 (approved):
👍🏼 Nice @rxu17 !  I think this looks right and I checked on AWS to see if the folder was created.  I'm going approve, but does this folder hiearchy look right here:
image

rxu17:
Yup, that's the one we are going with:
`s3://{cloudformation_bucket}/{namespace}/recover/src/...`

PR #22, opened 2023-03-10
[ETL-336] Separate out upload artifacts actions for develop vs prod env

rxu17:
**Purpose:** This PR is an extension of [ETL-336 Upload artifacts](https://github.com/Sage-Bionetworks/recover/pull/21).
This update uses environment variables and separate github action jobs by environment (develop vs prod) that allow the use of different AWS credentials IAM roles and cloudformation buckets based on environment.


.github/workflows/upload-and-deploy.yaml
```diff
@@ -48,21 +48,60 @@ jobs:
...
-          role-to-assume: arn:aws:iam::914833433684:role/sagebase-github-oidc-sage-bion-ProviderRolerecover-VTOKECAKJUBV
+          role-to-assume: ${{ vars.AWS_CREDENTIALS_IAM_ROLE }}
           role-session-name: GitHubActions-${{ github.repository_owner }}-${{ github.event.repository.name }}-${{ github.run_id }}
           role-duration-seconds: 1200
 
       - name: Setup sam
         uses: aws-actions/setup-sam@v2
 
       - name: Set namespace for non-default branch or for tag
-        if: github.ref_name != 'main'
```
thomasyu888, line 59: Hmm, so thinking about this more, I think we should keep this.   Is the rationale of not deploying the main branch to the development AWS account that any feature branches will be tested in the development account?

It would be nice if there was always an exact replica of the infrastructure that is in develop account as it is in prod account. I don't currently see how this would be created:

```
recover-dev-cloudformation/recover/recover...
```

.github/workflows/upload-and-deploy.yaml
```diff
@@ -48,21 +48,60 @@ jobs:
...
+        shell: bash
+        run: |
+          pipenv install --dev
+
+      - name: configure aws credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-region: us-east-1
+          role-to-assume: ${{ vars.AWS_CREDENTIALS_IAM_ROLE }}
+          role-session-name: GitHubActions-${{ github.repository_owner }}-${{ github.event.repository.name }}-${{ github.run_id }}
```
thomasyu888, line 100: Nit: I think eventually - we actually want to update this: https://github.com/Sage-Bionetworks/action-pipenv-aws-setup to use OIDC integration.  That may be cleaner than all the duplicated code we have.

Lets create a ticket for this?

thomasyu888 (changes requested):
Great initial PR, some things to discuss.

.github/workflows/upload-and-deploy.yaml
rxu17: Done: [IBCDPE-516](https://sagebionetworks.jira.com/browse/IBCDPE-516)

.github/workflows/upload-and-deploy.yaml
rxu17: I think we want to mirror our other buckets (parquet, intermediate) in that non-main branch deployments will always be namespaced in the dev AWS account because currently in the BridgeDownstream dev account, the cloudformation bucket's contents are not namespaced whilest the other bucket's contents are. I'm not sure of the original rationale but I would support the above of testing feature branches only in dev.

I think the structure should be the same, just that the infrastructure in the prod cloudformation bucket won't be namespaced:

- dev account: recover-dev-cloudformation/etl-416/recover/src/...
- prod account: recover-cloudformation/recover/src/

.github/workflows/upload-and-deploy.yaml
```diff
@@ -48,21 +48,60 @@ jobs:
...
         run: echo "NAMESPACE=$GITHUB_REF_NAME" >> $GITHUB_ENV
 
-      - name: Copy files to templates bucket, use dev cloudformation bucket if not main branch
-        if: github.ref_name != 'main'
-        run: python src/scripts/manage_artifacts/artifacts.py --upload --namespace $NAMESPACE --cfn_bucket recover-dev-cloudformation
+      - name: Copy files to templates bucket, use dev cloudformation bucket
+        run: python src/scripts/manage_artifacts/artifacts.py --upload --namespace $NAMESPACE --cfn_bucket ${{ vars.CFN_BUCKET }}
+
+
+  upload-files-main:
```
thomasyu888, line 65: Per our huddle, let's table the production deployment until we have the deployment in the test account completely figured out.  Please comment this out so we don't lose all the effort you put in.

.github/workflows/upload-and-deploy.yaml
```diff
@@ -48,21 +48,60 @@ jobs:
...
+    name: Upload files to S3 bucket in production
+    runs-on: ubuntu-latest
+    needs: pre-commit
+    if: github.ref_name == 'main'
+    # These permissions are needed to interact with GitHub's OIDC Token endpoint.
+    permissions:
+      id-token: write
+      contents: read
+    environment: prod
```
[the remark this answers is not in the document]
rxu17: Yes, most likely by using a [reusable workflow](https://docs.github.com/en/actions/using-workflows/reusing-workflows#creating-a-reusable-workflow) or updating the https://github.com/Sage-Bionetworks/action-pipenv-aws-setup

.github/workflows/upload-and-deploy.yaml
rxu17: I think I'm just going to update the environment of this `upload-files-main` to be `develop`. This is then saying that if the PR is good to go and gets merged into main, then the deployment of the main branch will upload artifacts to the develop account, guaranteeing us the latest version of the artifacts in the dev account bucket but with no namespace attached/it won't get deleted when cleanup happens as well in the future. Hope this is an okay resolution.

rxu17: Removed `upload-files-main` job block and instead added the updating of the non-namespaced `recover-dev-cloudformation/recover/src` and namespaced artifacts bucket to under the `upload-files` job block

thomasyu888 (approved):
🦖 LGTM! Thanks Rixing!

PR #24, opened 2023-03-16
[ETL-365] Launch stacks during push and PR to dev AWS account

rxu17:
**Purpose:** This PR follows the same logic as BridgeDownstream in that it launches stacks to the "recover" and feature branch namespaces of the recover **develop** account. 

This PR also parameterizes some of the python and glue versions used in the glue job stacks as we have varying python and glue versions to be used in the highest level stack config:`config/config.yaml`. Was unable to dynamically parameterize the lambda python runtime in the sam template to this stack because findings say it's not possible to pass this through to the lambda sam template.

- JSON to Parquet glue job: **Glue version 4.0** which supports **Python 3.10**
- S3 to Parquet glue job: **Python 3.9** (this is their highest available python for the glue python shell jobs) which means it can only use **Glue version 3.0**
- Lambda function for S3 to JSON: **Python 3.9** (this is their highest python version that they support, but 3.10 should be coming along in a couple of months)
- Pipenv python/overall environment: **Python 3.9** otherwise we can't sam build lambda especially when using GH actions

### DIFF
--- a/.github/workflows/upload-and-deploy.yaml
+++ b/.github/workflows/upload-and-deploy.yaml
@@ -27,6 +27,7 @@ jobs:
     permissions:
       id-token: write
       contents: read
+    environment: develop
     steps:
 
       - name: Setup code, pipenv, aws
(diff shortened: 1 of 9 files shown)


thomasyu888 (approved):
👽 LGTM!  This is a lot of combinations of dependencies.. Hopefully someday the dependencies are all supported by AWS...

PR #26, opened 2023-03-21
[ETL-384] Add production s3 ingestion bucket

thomasyu888:
Purpose: This creates the ingestion bucket for production data. (SYNAPSE_INPUT_SHARED_BUCKET)

This is the first bucket in the ingress pipeline for RECOVER project. We expect to receive data from Care Evolution in this bucket. This would be the raw version of the data, no QC, no QA - just all the data dump from Care Evolution. Both Care Evolution and Sage have access to this bucket

### DIFF
--- a/config/prod/s3-ingestion.yaml
+++ b/config/prod/s3-ingestion.yaml
@@ -8,6 +8,8 @@ stack_tags:
   # This is the shared account (credentials shared on Last Pass) for RECOVER
   CostCenter: "MGH RECOVER / 122500"
 parameters:
+  # (Optional) Name of the created bucket.
+  BucketName: "recover-ingestion-prod"
   # (Optional) true for read-write bucket, false (default) for read-only bucket
   AllowWriteBucket: 'true'
   # (Optional) Allow accounts, groups, and users to access bucket (default is no access).
--- a/config/prod/s3-ingestion.yaml
+++ b/config/prod/s3-ingestion.yaml
@@ -16,6 +16,3 @@ parameters:
     - 'arn:aws:iam::325565585839:root'   # Required ARN for a synapse bucket
     - 'arn:aws:sts::526515999252:assumed-role/AWSReservedSSO_S3ExternalCollab_40c062f682e7f3f5/meghasyam@sagebase.org'
     - 'arn:aws:iam::621233246578:role/EES3-RK-6A32F45B-RECOVER' # QA ARN role for Care Evolution
-sceptre_user_data:
-  SynapseIDs:
-    - "3357179" # Megha's synapse profileID
--- a/config/prod/s3-ingestion.yaml
+++ b/config/prod/s3-ingestion.yaml
@@ -1,16 +1,5 @@
-# Provision a general S3 bucket or a Synapse External Bucket
-# http://docs.synapse.org/articles/custom_storage_location.html
-
-# For use in RECOVER Synapse project (syn51105296)
-
-####
 # NOTE:
-# This is the second bucket in the ingress pipeline for RECOVER project. We will create a copy of the first bucket (recover_input_shared_bucket).
-# This bucket is to make sure we have a data source that is constant and cannot be edited by externals
-# Only Sage has access to this bucket, and this will be linked to the RECOVER Synapse project (syn51105296).
-# The file list will be replicated in Synapse via STS token
-#(https://help.synapse.org/docs/Compute-Directly-on-Data-in-Synapse-or-S3.2048426057.html#ComputeDirectlyonDatainSynapseorS3-Synapse-ManagedSTSStorageLocations)
-####
+# This is the first bucket in the ingress pipeline for RECOVER project.
 template:
   type: "http"
   url: "https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.5.0/templates/S3/synapse-external-bucket.j2"
--- a/config/prod/s3-ingestion.yaml
+++ b/config/prod/s3-ingestion.yaml
@@ -5,7 +5,6 @@ template:
   url: "https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.5.0/templates/S3/synapse-external-bucket.j2"
 stack_name: "recover-ingestion-prod"
 stack_tags:
-  OwnerEmail: 'synapse-service-sysbio-recover-data-storage-01@sagebase.org'
   # This is the shared account (credentials shared on Last Pass) for RECOVER
   CostCenter: "MGH RECOVER / 122500"
 parameters:
--- a/config/prod/s3-ingestion.yaml
+++ b/config/prod/s3-ingestion.yaml
@@ -3,13 +3,13 @@
 template:
   type: "http"
   url: "https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.5.0/templates/S3/synapse-external-bucket.j2"
-stack_name: "recover-ingestion-prod"
+stack_name: "recover-ingestion-bucket"
 stack_tags:
   # This is the shared account (credentials shared on Last Pass) for RECOVER
   CostCenter: "MGH RECOVER / 122500"
 parameters:
   # (Optional) Name of the created bucket.
-  BucketName: "recover-ingestion-prod"
+  BucketName: "recover-ingestion"
   # (Optional) true for read-write bucket, false (default) for read-only bucket
   AllowWriteBucket: 'true'
   # (Optional) Allow accounts, groups, and users to access bucket (default is no access).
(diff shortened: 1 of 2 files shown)


config/prod/s3-ingestion.yaml
```diff
@@ -0,0 +1,21 @@
+# NOTE:
+# This is the first bucket in the ingress pipeline for RECOVER project.
+template:
+  type: "http"
+  url: "https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.5.0/templates/S3/synapse-external-bucket.j2"
```
rxu17, line 5: Since we're not linking the ingestion bucket to synapse, do we need to still use this external template/all of the below synapse related configurations or should we just be updating our [recover s3 template yaml](https://github.com/Sage-Bionetworks/recover/blob/main/templates/s3-bucket.yaml)'s bucket policy with the necessary permissions to grant to these specific accounts?

thomasyu888: Thanks @rxu17 for the quick review.  So a couple things - I removed the sceptre_user_data, but.

1.  I'm not sure our bucket template is set up to Grant access to: 
  
    ```   
    - 'arn:aws:sts::526515999252:assumed-role/AWSReservedSSO_S3ExternalCollab_40c062f682e7f3f5/meghasyam@sagebase.org'
    - 'arn:aws:iam::621233246578:role/EES3-RK-6A32F45B-RECOVER' # QA ARN role for Care Evolution
    ```

1. That said we may be able to remove the role for Megha down the line when we grant him access to the RECOVER prod aws account as a developer.

config/prod/s3-ingestion.yaml
```diff
@@ -0,0 +1,18 @@
...
+  # This is the shared account (credentials shared on Last Pass) for RECOVER
+  CostCenter: "MGH RECOVER / 122500"
+parameters:
+  # (Optional) true for read-write bucket, false (default) for read-only bucket
+  AllowWriteBucket: 'true'
+  # (Optional) Allow accounts, groups, and users to access bucket (default is no access).
+  GrantAccess:
+    - 'arn:aws:iam::325565585839:root'   # Required ARN for a synapse bucket
+    - 'arn:aws:sts::526515999252:assumed-role/AWSReservedSSO_S3ExternalCollab_40c062f682e7f3f5/meghasyam@sagebase.org'
+    - 'arn:aws:iam::621233246578:role/EES3-RK-6A32F45B-RECOVER' # QA ARN role for Care Evolution
```
rxu17, line 18: I think we need a `BucketName` parameter as mentioned here: https://github.com/Sage-Bionetworks/aws-infra/blob/v0.5.0/templates/S3/synapse-external-bucket.j2#L64-L67

config/prod/s3-ingestion.yaml
```diff
@@ -0,0 +1,18 @@
+# NOTE:
+# This is the first bucket in the ingress pipeline for RECOVER project.
+template:
+  type: "http"
+  url: "https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.5.0/templates/S3/synapse-external-bucket.j2"
+stack_name: "recover-ingestion-prod"
+stack_tags:
+  OwnerEmail: 'synapse-service-sysbio-recover-data-storage-01@sagebase.org'
```
rxu17, line 8: I think based on this comment: https://github.com/Sage-Bionetworks/scicomp-provisioner/pull/566#discussion_r1142612670
and also from the template, `CostCenter` and `OwnerEmail` aren't available in the template and should be removed

rxu17 (commented):

config/prod/s3-ingestion.yaml
rxu17: It's not, we'd need to modify our bucket template. 
After looking through the external template further, it sounds good to me/makes sense to just use that!

rxu17:
The only thing I'm not sure about is if this [synapse-external-bucket.j2#L84-L90](https://github.com/Sage-Bionetworks/aws-infra/blob/v0.5.0/templates/S3/synapse-external-bucket.j2#L84-L90) means it will block all public access? We had to set something up like this for our buckets in recover to block public access: https://github.com/Sage-Bionetworks/recover/pull/15/files

config/prod/s3-ingestion.yaml
thomasyu888: Great idea!

config/prod/s3-ingestion.yaml
thomasyu888: Excellent - I removed `OwnerEmail` but `CostCenter is still required

config/prod/s3-ingestion.yaml
```diff
@@ -0,0 +1,18 @@
...
+  # This is the shared account (credentials shared on Last Pass) for RECOVER
+  CostCenter: "MGH RECOVER / 122500"
+parameters:
+  # (Optional) true for read-write bucket, false (default) for read-only bucket
+  AllowWriteBucket: 'true'
+  # (Optional) Allow accounts, groups, and users to access bucket (default is no access).
+  GrantAccess:
+    - 'arn:aws:iam::325565585839:root'   # Required ARN for a synapse bucket
+    - 'arn:aws:sts::526515999252:assumed-role/AWSReservedSSO_S3ExternalCollab_40c062f682e7f3f5/meghasyam@sagebase.org'
```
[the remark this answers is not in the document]
thomasyu888: great question @rxu17, could you help me ask IT this question?

config/prod/s3-ingestion.yaml
```diff
@@ -0,0 +1,19 @@
+# NOTE:
+# This is the first bucket in the ingress pipeline for RECOVER project.
+template:
+  type: "http"
+  url: "https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.5.0/templates/S3/synapse-external-bucket.j2"
+stack_name: "recover-ingestion-prod"
```
rxu17, line 6: Nit: Could we for consistency, could we follow recover's pattern for naming stacks and bucket_names (see example for prod buckets): https://github.com/Sage-Bionetworks/recover/blob/main/config/prod/s3-processed-data-bucket.yaml#L4,L8

Something like:
`BucketName: "recover-ingestion"`
`stack_name: "recover-ingestion-bucket"`

config/prod/s3-ingestion.yaml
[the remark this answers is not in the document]
rxu17: Per this discussion (here)[https://sagebionetworks.slack.com/archives/CEFQD0KU1/p1679430261246839], it doesn't seem like the template would block public access automatically. 
We can either 

- wait until April 2023 (not sure what date) when the bucket created initially will have the public access automatically blocked
- Just create the bucket now, and then modify the public bucket policy on the AWS console
- Request update to the external j2 template to add the configurations needed to block public access
- Update this to use our [recover s3 template yaml](https://github.com/Sage-Bionetworks/recover/blob/main/templates/s3-bucket.yaml) instead, and update our yaml to do all the things we want it to do such as grant access, etc.

config/prod/s3-ingestion.yaml
thomasyu888, line 6:
```suggestion
stack_name: "recover-ingestion"
```

config/prod/s3-ingestion.yaml
```diff
@@ -0,0 +1,19 @@
...
+stack_tags:
+  # This is the shared account (credentials shared on Last Pass) for RECOVER
+  CostCenter: "MGH RECOVER / 122500"
+parameters:
+  # (Optional) Name of the created bucket.
+  BucketName: "recover-ingestion-prod"
```
thomasyu888, line 12:
```suggestion
  BucketName: "recover-ingestion"
```

config/prod/s3-ingestion.yaml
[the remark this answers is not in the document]
thomasyu888: @rxu17 Is this true: https://sagebionetworks.slack.com/archives/CEFQD0KU1/p1679437439776059?thread_ts=1679430261.246839&cid=CEFQD0KU1.  It seems we don't need to specify it.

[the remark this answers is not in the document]
rxu17: From my experience, it only disables the ACL specific settings under the public access portion. We had to do this: https://github.com/Sage-Bionetworks/recover/pull/15/files even after our bucket had the [ObjectOwnership: BucketOwnerEnforced](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-ownershipcontrolsrule.html) set to block all of the public access points.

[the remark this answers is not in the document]
rxu17: I agree with the quickest way

rxu17 (approved):
I approve this.

PR #29, opened 2023-03-29
[ETL-398] Add glue, python version params to relevant prod glue configs

rxu17:
**Purpose**: Prod glue job stack configs need to be updated to use the new glue and python version parameters that dev glue job stack configs use.

This also **blocks** a lot of tickets that involve testing stack deployment in prod because currently the glue job stacks error out without this update


thomasyu888 (approved):
🚀 Great! LGTM

PR #31, opened 2023-03-30
[ETL-395] Modify prod stack configs to use new pre-ETL bucket name 

rxu17:
**Purpose:** We recently created the new pre-ETL (input data) in the prod account, and now the prod stack configs need to be adjusted to use the new bucket. Here a stack group config was also created to house the prod specific bucket names for ease of use across all the prod stacks.

### DIFF
--- a/config/prod/config.yaml
+++ b/config/prod/config.yaml
@@ -1,0 +1,5 @@
+ingestion_bucket_name: recover-ingestion
+input_bucket_name: recover-input-data
+cloudformation_artifact_bucket_name: recover-cloudformation
+intermediate_bucket_name: recover-intermediate-data
+processed_data_bucket_name: recover-processed-data
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -2,7 +2,6 @@ project_code: recover
 namespace: {{ var.namespace | default('recover') }}
 region: us-east-1
 synapseAuthSsmParameterName: synapse-recover-auth
-source_bucket: recover-s3-bucket-bucket-1rb2h1xfnxpoi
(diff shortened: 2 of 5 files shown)


config/prod/config.yaml
```diff
@@ -0,0 +1,5 @@
+ingestion_bucket_name: recover-ingestion
+input_bucket_name: recover-input-data
+cloudformation_artifact_bucket_name: recover-cloudformation
+intermediate_bucket_name: recover-intermediate-data
+processed_data_bucket_name: recover-processed-data
```
philerooski, line 5: Instead of including the bucket names in the stack group config we should list the bucket stacks as dependencies in the relevant config files and reference the bucket name using 
```
!stack_output_external bucket-stack::BucketName
```

rxu17: Based on our recent slack discussion, and to avoid redundancy in our stack dependencies with the buckets, we will use the `stack_group_config` way

philerooski: You'll also want to use a `stack_group_config` reference for the BucketName parameter in the bucket stacks themselves: `config/prod/s3-cloudformation-bucket.yaml`, `config/prod/s3-ingestion.yaml`, etc.

philerooski: Feel free to do the same in dev.

rxu17: Yup, I have a ticket that encompasses all of that here: https://sagebionetworks.jira.com/browse/ETL-401

philerooski (approved):

PR #32, opened 2023-03-30
[ETL-386] Rename pilot data bucket and link to Synapse

philerooski:
* Pilot bucket stack renamed to recover-dev-input-bucket and bucket renamed to recover-dev-input-data
* Synapse permissions granted through use of `SynapseIds` parameter.

### DIFF
--- a/config/develop/glue-job-role.yaml
+++ b/config/develop/glue-job-role.yaml
@@ -2,7 +2,7 @@ template:
   path: glue-job-role.yaml
 stack_name: glue-job-role
 parameters:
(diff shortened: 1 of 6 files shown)


thomasyu888:
Thanks @philerooski , can we add a short description in the PR and look into why the CI/CD is failling?  Thanks!

rxu17:
I'm thinking this and [ETL-396](https://sagebionetworks.jira.com/browse/ETL-396) will have to be combined into one ticket in order for CI/CD to go through otherwise, can merge this first and then follow up with ETL-396 which should resolve the CI/CD.

thomasyu888 (approved):
🔥🔥🔥

PR #33, opened 2023-03-30
[ETL-394] Update prod and dev account resources to use NAMESPACE

rxu17:
**Purpose:** This updates the prod stack configs to use **namespaced** files and stack names to mirror dev stack configs. This also organizes the stacks into namespaced folders and adds a new namespace feature for both dev and prod stack configs for the S3 to JSON job. 

> One point of discussion is if we should namespace JSON to parquet jobs as well (including the individual jobs) similar to BridgeDownstream or leave it like it is for recover since we'll only have one "study" (how we decide on staging production data might affect this later down the line)?


philerooski (approved):
I'm not much of a stickler for perfectly clean commit logs, but 8 commits is a bit extreme for a small change like this. Can you [git squash](https://www.git-tower.com/learn/git/faq/git-squash) everything into a single commit?

thomasyu888:
@rxu17 , @philerooski you can 
image

Also did we want to have that discussion in the production testing jira ticket?

PR #34, opened 2023-03-30
[ETL-382] Update folder structure of dev resources

rxu17:
**Purpose:** This PR corrects the filepaths that various configs point to for their scripts so it follows this pattern now:
`<namespace>/src/...`. Our default namespace will be `main` when deploying to main branch as well as in production and CI/CD has been updated to reflect that.


thomasyu888 (approved):
💯 Thanks!  This looks great!

PR #35, opened 2023-03-31
[ETL-401] Update dependencies for stack configs in prod and dev accounts

rxu17:
**Purpose:** This PR updates and syncs up the dependencies in the prod and dev stack configs. 
This includes:

- Adding required dependencies where needed
- Removing dependencies pointing to buckets and instead referencing bucket names using `stack_group_config` to reduce redundancy and spinning up resources when unnecessary

### DIFF
--- a/config/develop/config.yaml
+++ b/config/develop/config.yaml
@@ -1,0 +1,5 @@
+ingestion_bucket_name: recover-dev-ingestion
+input_bucket_name: recover-dev-input-data
+cloudformation_artifact_bucket_name: recover-dev-cloudformation
+intermediate_bucket_name: recover-dev-intermediate-data
+processed_data_bucket_name: recover-dev-processed-data
--- a/config/develop/glue-job-JSONToParquet.yaml
+++ b/config/develop/glue-job-JSONToParquet.yaml
@@ -1,15 +1,13 @@
 template:
   path: glue-job-JSONToParquet.j2
 dependencies:
-  - develop/s3-processed-data-bucket.yaml
-  - develop/s3-cloudformation-bucket.yaml
   - develop/glue-job-role.yaml
 stack_name: glue-job-JSONToParquet
 parameters:
   JobDescription: Export JSON datasets to Parquet
   JobRole: !stack_output_external glue-job-role::RoleArn
-  TempS3Bucket: !stack_output_external recover-dev-processed-data-bucket::BucketName
-  S3ScriptBucket: !stack_output_external recover-dev-cloudformation-bucket::BucketName
+  TempS3Bucket: {{ stack_group_config.processed_data_bucket_name }}
+  S3ScriptBucket: {{ stack_group_config.cloudformation_artifact_bucket_name }}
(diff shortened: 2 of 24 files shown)


config/develop/glue-job-JSONToParquet.yaml
```diff
@@ -1,15 +1,13 @@
 template:
   path: glue-job-JSONToParquet.j2
 dependencies:
-  - develop/s3-processed-data-bucket.yaml
```
thomasyu888, line 4: So, did this exist because this job would fail without these buckets?

config/develop/namespaced/glue-workflow.yaml
```diff
@@ -3,10 +3,11 @@ template:
 stack_name: "{{ stack_group_config.namespace }}-glue-workflow"
 dependencies:
   - develop/namespaced/glue-tables.yaml
+  - develop/namespaced/glue-job-S3ToJsonS3.yaml
```
thomasyu888, line 6: Can you explain why you added this?

config/develop/config.yaml
thomasyu888, line 1: I might have missed this, but does `config/prod/config.yaml` exist already?

thomasyu888 (commented):
Nice!  Just some had some questions prior to approving.

config/develop/config.yaml
rxu17: Yup!

config/develop/namespaced/glue-workflow.yaml
rxu17: The glue workflow needs the S3 to JSON job name to create the S3 to JSON job trigger [here] (https://github.com/Sage-Bionetworks/recover/blob/main/templates/glue-workflow.j2#L69) so the S3 to JSON job stack needs to be created before this `glue-workflow` stack gets created. Before, deployment was failing because it was trying to launch the `glue-workflow` stack but it couldn't find the S3 to JSON job stack to get the name.

config/develop/glue-job-JSONToParquet.yaml
rxu17: Hmmm my thought process behind this is generally we'd want to create all of our buckets and leave them alone prior to running our pipelines I believe so they would all exist.

The removal of the buckets as dependencies and instead relying on referencing it through the `{{ stack_group_config.input_bucket_name }}` is to just reduce redundancy because a lot of the stacks use the buckets as dep. One example is that `cfn-s3objects-macro` won’t run every time we, for example, update the lambda stack because the input bucket is not a dependency for the lambda stack anymore. And if there are any updates to the buckets themselves they would be updated anyways with our CI/CD..

philerooski: We don't need the bucket stack to deploy this resource. If the buckets didn't exist - yes - the `TempS3Bucket` and `S3ScriptBucket` parameters would point to a non-existent location and the job would fail if we tried to run it. But there's nothing preventing us from creating the job before the buckets come into existence.

philerooski (approved):

thomasyu888 (approved):
🚀 LGTM!

PR #37, opened 2023-04-03
[ETL-381] Grant Synapse access to dev parquet bucket

philerooski:
* Synapse granted permissions to dev parquet bucket (by modifying existing bucket stack) and owner.txt stack created
* Added functionality to `setup_external_storage` script to be able to use Synapse authentication token stored in SSM parameter.
* I tested it on this Synapse folder: https://www.synapse.org/#!Synapse:syn51284492

### DIFF
--- a/config/develop/s3-processed-data-bucket-owner-txt.yaml
+++ b/config/develop/s3-processed-data-bucket-owner-txt.yaml
@@ -1,0 +1,13 @@
+template:
+  type: file
+  path: s3-owner-txt.yaml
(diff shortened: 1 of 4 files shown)


rxu17 (approved):
LGTM!

thomasyu888 (approved):
🚀 LGTM - Thanks for testing this!  we can definitely give this script to people on SysBio when they need to link the pre-ETL bucket to a project.  That said, they probably have a script already.

PR #27, opened 2023-03-21
[ETL-385] Add in s3 production data pre-etl bucket

thomasyu888:
Purpose: This PR create the production data pre-etl bucket which the ETL pipeline will read from.

This bucket is to make sure we have a data source that is constant and cannot be edited by externals. Only Sage has access to this bucket, and this will be linked to the RECOVER Synapse project (syn51105296). The file list will be replicated in Synapse via STS token

PR #38, opened 2023-04-05
[ETL-385] Grant prod input bucket synapse access

rxu17:
**Purpose:** Minor adjustment to add the ability to connect to Synapse and provide STS access to mirror the dev input bucket as well. Followup of this [PR](https://github.com/Sage-Bionetworks/recover/pull/27).


thomasyu888 (approved):
🚀 Thanks!!

PR #39, opened 2023-04-05
[ETL-410] Add namespace for json to parquet and glue database stacks

rxu17:
**Purpose:** Namespace JSON to parquet stack and jobs in order to allow for separate feature branch development. Add some namespacing for the glue database to allow for easier monitoring of the glue tables in the console.


thomasyu888 (approved):
🌳 Thanks @rxu17 , I agree with this change, but could you assist me in understanding why this makes it easier to monitor the glue tables in the console?

Going to pre-approve!

philerooski (approved):

rxu17:
> 🌳 Thanks @rxu17 , I agree with this change, but could you assist me in understanding why this makes it easier to monitor the glue tables in the console?
> 
> Going to pre-approve!
If you look at this view (glue table view) which you can see [here](https://us-east-1.console.aws.amazon.com/glue/home?region=us-east-1#/v2/data-catalog/tables), you can see that in the **Database** column, we have etl-410=gluedatabase and then the rest of the entries are some random names. So now you can filter by your feature branch name vs before you probably had to look at the filepath column (not in this pic) which is also harder to filter by.

image

PR #40, opened 2023-04-06
[ETL-378] Launch all stacks to "staging" namespace in production account

rxu17:
**Purpose:** Currently this github action workflow will deploy to the "staging" namespace of production upon merge. If we decide we want to switch out of the staging testing workflow, it will be an easy change to have it deploy to the "main" namespace of prod. I also had the deployment to production as a separate job in case we wanted to do anything additional with that GH action job in the future.

Will squash the commits upon merge.

### DIFF
--- a/.github/workflows/upload-and-deploy.yaml
+++ b/.github/workflows/upload-and-deploy.yaml
@@ -48,7 +48,6 @@ jobs:
 
   sceptre-deploy-develop:
     name: Deploys branch using sceptre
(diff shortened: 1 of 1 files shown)


thomasyu888 (approved):
🏁 Great!  I posted on the Jira ticket about this, but eventually, we will want to think about automatic non-staging production deployment.  Maybe via git release or some other mechanism.

philerooski (approved):

PR #41, opened 2023-04-14
[ETL-289] Remove deleted and duplicated samples

philerooski:
This PR covers removing deleted samples. (Removing duplicate samples was already implemented).

If a data type is a HealthKit data type, we check if it has a respective "deleted" table. If it does, we remove records which have a sample key in the "deleted" table. Otherwise, we carry on as usual without removing any records.

I also did some minor clean up in `templates/glue-job-role.yaml` and fixed a bug in `templates/glue-workflow.j2`


thomasyu888 (approved):
🚀 Nice!

PR #42, opened 2023-04-14
[ETL-409] Create python comparison script

rxu17:
Hello! The purpose of this script is to compare the parquet datasets by data type in the established "main" namespace and the new "staging" namespace of the processed data (post-ETL) bucket. 

This is just a draft PR because tests are still in progress/need some troubleshooting/should get grouped into classes for better organization but most/all of the code and functionality is complete. However would love to get some feedback/organizational ideas (for example: I feel like some of the functions could just be moved into a common utils module but at the same time if they never need to get used again in this repo/if they ever plan to be refactored, it might be harder to do that then. This PR also unfortunately got pretty big because there were a bunch of edge cases and logic in regards to the flow of to consider when comparing between two datasets.  

This code contains the following **main functionality:**

- Compares the data type folders present in each namespace 

Does the following comparisons by dataset data type (e.g: dataset_fitbitactivitylogs) **ONLY** if the datasets meet the following conditions:

- Neither dataset is empty
- The datasets should have columns in common to compare as a chunk of the comparisons relies on that

Comparisons done:
- Compares the two datasets' column data types
- Compares the two datasets' row count
- Compares the two datasets' column names
- Compares the two datasets' column values (once they are the same dim)
- Compares the two datasets' row values (once they are the same dim)

Added some comments about thoughts on certain code. Also any glue related code will be added as a part of [ETL-406](https://sagebionetworks.jira.com/browse/ETL-406)

[ETL-406]: https://sagebionetworks.jira.com/browse/ETL-406?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/glue/jobs/compare_parquet_datasets.py
```diff
@@ -0,0 +1,422 @@
...
+import synapseclient
+from pyarrow import fs
+import pyarrow.parquet as pq
+
+# from json_to_parquet import INDEX_FIELD_MAP
+
+logger = logging.getLogger()
+logger.setLevel(logging.INFO)
+
+INDEX_FIELD_MAP = {
```
rxu17, line 17: Note that this will get removed if we decide to move this `INDEX_FIELD_MAP` to a common script or just use the one in `json_to_parquet` but for ease of testing (because would have to do it in a docker image due to json_to_parquet requiring glue), it seemed easier to leave this in.

src/glue/jobs/compare_parquet_datasets.py
```diff
@@ -0,0 +1,422 @@
...
+            content["Key"].split("/")[-1]
+            for content in contents
+            if content["Key"].split("/")[-1] != "owner.txt"
+        ]
+    else:
+        folders = []
+    return folders
+
+
+def keep_common_rows_cols(
```
rxu17, line 167: I'm also not sure about this method I used to get the two datasets to the same dim (combination of subsetting, merge and then separation) so we can use something like `pandas.compare`. This method is used in: `compare_dataset_row_vals`

src/glue/jobs/compare_parquet_datasets.py
```diff
@@ -0,0 +1,422 @@
...
+    else:
+        # check that the dataset has no dup cols or dup rows and that they have cols in common
+        comparison_result[data_type] = []
+        # check if one or both of the datasets have no data
+        if staging_dataset.empty or main_dataset.empty:
+            comparison_result["empty"][
+                data_type
+            ] = f"One of {args.staging_namespace} or {args.main_namespace} has no data. Comparison cannot continue."
+        else:
+            comparison_result[data_type].append(
```
rxu17, line 376: Still have to organize the structuring of the comparison messages within the dict. Trying to mirror/get inspiration from what we do in BridgeDownstream with the validate schema stuff.

src/glue/jobs/compare_parquet_datasets.py
philerooski: This redundancy is fine. I don't think it's as simple to remove as including a config.py with this constant in the same directory, since these are glue job scripts that are stored in S3.

tests/test_compare_parquet_datasets.py
```diff
@@ -0,0 +1,355 @@
...
+    file_key = "staging/parquet/dataset_fitbitactivitylogs/test.parquet"
+    parquet_dataset = compare_parquet.get_parquet_dataset(
+        dataset_key=f"{parquet_bucket_name}/{file_key}",
+        s3_filesystem=mock_s3_filesystem,
+    )
+    assert parquet_dataset == None
+
+
+@mock_s3
+def test_that_get_parquet_dataset_returns_dataset_if_datasets_exist(
```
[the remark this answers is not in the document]
philerooski: You might need to write a mocked class rather than use moto.

src/glue/jobs/compare_parquet_datasets.py
rxu17: This will be resolved with implementation of [datacompy](https://github.com/capitalone/datacompy)

src/glue/jobs/compare_parquet_datasets.py
rxu17: This will be mostly resolved with [datacompy](https://github.com/capitalone/datacompy)/follow its lead

Pipfile
```diff
@@ -13,3 +13,5 @@ pre-commit = "*"
 sceptre = ">=3.2.0"
 sceptre-sam-handler = "*"
 synapseclient = "~=2.7"
+pandas = "*"
+moto = "*"
```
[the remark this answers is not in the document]
thomasyu888: So for the python client 2.7.1, we can only do pandas < 1.5 right now.

thomasyu888 (changes requested):
Excellent work! Please see comments.

tests/test_compare_parquet_datasets.py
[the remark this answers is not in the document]
rxu17: Sadly it seems like specifically pyarrow's filesystem can't be mocked because It seems that using mock for testing purposes fails, as pyarrow (not in the most Pythonic way) checks for the type of the filesystem parameter passed to write_to_dataset: https://github.com/apache/arrow/blob/main/python/pyarrow/filesystem.py#L443-L459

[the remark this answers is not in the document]
rxu17: > 

So one way to approach this is to do something like:

patch.object(pq, "read_table"...)

What you care about is that a dataframe is returned from this function.

Yup, that is the plan

tests/test_compare_parquet_datasets.py
rxu17, line 1: We have over 20 tests and I'm wondering for organization if it would be worthwhile to group the tests into classes instead (the grouping into individual testing module method seems overkill)? A class for each function's tests seems overkill so maybe it would be more like a class for each category (e.g the tests that test the missing, additional and common columns could be grouped into one test). 

Similarly with the tests surrounding the s3 folders could be another class.

But stand-alone functions' tests would probably not get grouped into classes

src/glue/jobs/compare_parquet_datasets.py
```diff
@@ -0,0 +1,474 @@
...
+
+        comparison_report = add_additional_msg_to_comparison_report(
+            comparison_report,
+            add_msgs=compare_column_names(data_type, staging_dataset, main_dataset),
+            msg_type="column_name_diff",
+        )
+    return f"{header_msg}{comparison_report}"
+
+
+def main():
```
rxu17, line 407: At this point, I'm thinking this main function's logic (just this one if-else statement) would just have to be tested with some test data rather than writing any test functions just because it would be hard to even compare the results since it's a large chunk of string. Every function here in `main` is tested in their own tests.

src/glue/jobs/compare_parquet_datasets.py
```diff
@@ -0,0 +1,473 @@
...
+        abs_tol=0,  # Optional, defaults to 0
+        rel_tol=0,  # Optional, defaults to 0
+        df1_name=staging_namespace,  # Optional, defaults to 'df1'
+        df2_name=main_namespace,  # Optional, defaults to 'df2'
+    )
+    compare.matches(ignore_extra_columns=False)
+    return compare.report()
+
+
+def add_additional_msg_to_comparison_report(
```
rxu17, line 278: All the string formatting messages scattered about are not pretty/ideal but tried to group/limit them to just this function and `compare_datasets_by_data_type`.

src/glue/jobs/compare_parquet_datasets.py
```diff
@@ -0,0 +1,473 @@
...
+
+    Returns:
+        str: the value as is
+    """
+    if value == "":
+        raise argparse.ArgumentTypeError('Argument value cannot be an empty string')
+    return value
+
+
+def get_s3_file_key_for_comparison_results(
```
rxu17, line 87: Are we alright with this format here when saving the comparison reports?

src/glue/jobs/compare_parquet_datasets.py
thomasyu888: I think this is great!

src/glue/jobs/compare_parquet_datasets.py
thomasyu888: I agree with your assessment.  An integration test is most likely easiest here instead of trying to mock. (1st option)

tests/test_compare_parquet_datasets.py
thomasyu888: Hmm, theres not really a right or wrong answer here.  I'm just also thinking of the other benefits that classes provide us other than grouping like re-using class attributes.  We use fixtures to setup and teardown, so thats not really an issue.  What I personally dislike about using a class to group tests by functionality is that you end up testing different functions within a class.

There are also methods like this for grouping: https://www.tutorialspoint.com/pytest/pytest_grouping_the_tests.htm.

Overall, I leave it to you.  I'm also ok as is - it might just take some time to figure out what tests need to be added/modified/removed in the future.

src/glue/jobs/compare_parquet_datasets.py
rxu17: Sounds good, just noting here only quirk I noticed about its behavior is that it's not happy with slashes if you have something like:

-  `os.path.join("s3://", "some_bucket_name", "some_folder", "/")` which turns the entire path into `/`
- `os.path.join("s3://", "/some_bucket_name", "some_folder")` which turns the path into `/some_bucket_name/some_folder`

philerooski: `s3://` is not technically part of the path. That's the scheme of the URI. So you can do something like:
```
f"s3://{os.path.join(bucket, key_prefix)}"
```

src/glue/jobs/compare_parquet_datasets.py
```diff
@@ -0,0 +1,473 @@
...
+        help=("The name of the main namespace to use"),
+        default="main"
+    )
+    parser.add_argument(
+        "--parquet-bucket",
+        required=True,
+        type = validate_args,
+        help=("The name of the S3 bucket containing the S3 files to compare"),
+        default="recover-dev-processed_data"
+    )
```
philerooski, line 65: This script assumes a lot of things about where the parquet datasets are located in S3. I think it's fine to leave it as it is now (we don't yet have a use case for needing a more general parameterization and it looks like a hassle to refactor), but I do wish that this script would have been able to accept arbitrary S3 locations of two parquet datasets.

philerooski (approved):
It's hard to tell by looking at the code if this will perfectly suit our purposes. Since you are most familiar with the comparisons being done here, can you think of any situation where we might inadvertently change data and the change isn't logged by this script?

For example, datacompy will only output the first 10 rows that differ in value. Perhaps these are expected changes, but rows 11 through whatever that differ in value will not be logged, and these could be unexpected changes (potentially in a different column than the expected changes). If we only merge in the smallest possible change that should minimize or even eliminate any risk of multi-column differences, but I'm curious if you can come up with any convoluted scenarios.

rxu17:
> It's hard to tell by looking at the code if this will perfectly suit our purposes. Since you are most familiar with the comparisons being done here, can you think of any situation where we might inadvertently change data and the change isn't logged by this script?
> 
> For example, datacompy will only output the first 10 rows that differ in value. Perhaps these are expected changes, but rows 11 through whatever that differ in value will not be logged, and these could be unexpected changes (potentially in a different column than the expected changes). If we only merge in the smallest possible change that should minimize or even eliminate any risk of multi-column differences, but I'm curious if you can come up with any convoluted scenarios.

The nice thing about this comparison is that it will print out every column that has differences (between the two datasets) but will give a high level summary regarding it (See [ETL-409](https://sagebionetworks.jira.com/browse/ETL-409) for more info). So even if the 10 rows won't capture all the differences, you can still see it in the column summary section is my conclusion of testing this resource. That being said, I'd imagine if we do get a bunch of unexpected differences that come up, we'd have to dig deeper into the dataset itself unless that is something we want covered by this tool?

Hmmm when testing this tool, I tried to come up with various possible dataset change scenarios (see conftest.py). Some other things I can think of is:

- We do read in the parquet datasets from a location and convert it to pandas DF prior to any comparison. If there is any unexpected behavior by that process, that would probably not be captured by this change log just because we do that to both the "staging" and "main" namespace datasets. Or it could make it seem like the "staging" datasets had no change because of that unexpected behavior. I think this can be mostly avoided by running a series of test cases on that `get_parquet_dataset` function
- I'm not too familiar with parquet `read_table` and if it ignores files in a location that just doesn't have `.parquet` in the file extension or also tries to read them in? There isn't a check to make sure the files we read in is equal to the ones we actually have in the folder. If this happens to both "staging" and "main", we wouldn't be able to tell as the comparisons would give us no differences in that area.
- I'll have to think more on this but this is what i have so far

src/glue/jobs/compare_parquet_datasets.py
thomasyu888: I added a ticket to track this ^

rxu17:
> It's hard to tell by looking at the code if this will perfectly suit our purposes. Since you are most familiar with the comparisons being done here, can you think of any situation where we might inadvertently change data and the change isn't logged by this script?
> 
> For example, datacompy will only output the first 10 rows that differ in value. Perhaps these are expected changes, but rows 11 through whatever that differ in value will not be logged, and these could be unexpected changes (potentially in a different column than the expected changes). If we only merge in the smallest possible change that should minimize or even eliminate any risk of multi-column differences, but I'm curious if you can come up with any convoluted scenarios.

Adding on to this. When there are missing rows/additional rows that are more than 10 rows, it may be the scenario that that we see a pattern in the first 10 rows, but without the rest of the rows printed out, it would be difficult to see if they were expected as well. 

The package is flexible in that we can set how many rows we can have the log print out. Since we are saving it as a txt file in the S3 location, we could just have it print out all of them. And given that it's a class, we have access to all of its individual methods (e.g: the method that compares two columns) so if we wanted to expand on them or use ones separately, we could as well. Now the question is do we want to save that for a future iteration or add those features now.

rxu17:
Added functionality to save csv files of all of the rows that are different between the staging and main datasets in the same area as the high level csv. Similarly with the column values are different as well. 

You can view an example run at:

- `recover-dev-processed-data/etl-409/comparison_result/dataset_healthkitv2samples__parquet_compare.txt` 
- `recover-dev-processed-data/etl-409/comparison_result/dataset_healthkitv2samples_all_diff_main_rows.csv`

tests/test_compare_parquet_datasets.py
rxu17: Sounds good, I'm going to add a ticket to make this code more efficient (which will include grouping the tests) because we may need more time to decide which method is best before rewriting everything.

src/glue/jobs/compare_parquet_datasets.py
```diff
@@ -0,0 +1,391 @@
...
+def has_common_cols(staging_dataset: pd.DataFrame, main_dataset: pd.DataFrame) -> list:
+    """Gets the list of common columns between two dataframes"""
+    common_cols = staging_dataset.columns.intersection(main_dataset.columns).tolist()
+    return common_cols != []
+
+
+def get_missing_cols(staging_dataset: pd.DataFrame, main_dataset: pd.DataFrame) -> list:
+    """Gets the list of missing columns present in main but not in staging"""
+    missing_cols = main_dataset.columns.difference(staging_dataset.columns).tolist()
+    return missing_cols
```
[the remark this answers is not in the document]
rxu17: Just yesterday, I found that datacompy.Compare has functions that handles some of the above (e.g: gets unique columns in one dataset but not the other here: https://github.com/capitalone/datacompy/blob/develop/datacompy/core.py#L220-L230. It's marked as a TODO because I feel like this PR has blown up already so much and having those changes is a bit more involved.

thomasyu888 (approved):
🦖 LGTM!  Tremendous effort!

thomasyu888:
@rxu17 I am ok with merging this and creating a separate ticket to track other work.

PR #43, opened 2023-04-18
[ETL-406] Add additional staging resources to production stack group

rxu17:
**Purpose:** I created this draft PR before getting into more nitty gritty edits/testing as it's waiting on [ETL-409](https://sagebionetworks.jira.com/browse/ETL-409) to get some feedback about the structuring of our cloudformation resources and if there were additional specs we expected from the two resources to be added as part of this ticket. I have added the following to both our development and production stacks:

-  compare parquet glue job
-  compare parquet glue trigger (triggers the above job)

I added the compare parquet glue trigger to our current glue-workflow instead of creating it as an additional resource. Once the compare parquet python script is complete/merged, should be able to properly test this.

### DIFF
--- a/config/develop/namespaced/glue-workflow.yaml
+++ b/config/develop/namespaced/glue-workflow.yaml
@@ -10,6 +10,9 @@ parameters:
   ParquetBucketName: {{ stack_group_config.processed_data_bucket_name }}
   GlueDatabase: !stack_output_external "{{ stack_group_config.namespace }}-glue-tables::DatabaseName"
   S3ToJsonJobName: !stack_output_external "{{ stack_group_config.namespace }}-glue-job-S3ToJsonS3::JobName"
+  CompareParquetJobName: !stack_output_external "{{ stack_group_config.namespace }}-glue-job-CompareParquet::JobName"
+  CompareParquetStagingNamespace: {{ stack_group_config.namespace }}
+  CompareParquetMainNamespace: "main"
 stack_tags:
   {{ stack_group_config.default_stack_tags }}
 sceptre_user_data:
--- a/config/develop/namespaced/glue-workflow.yaml
+++ b/config/develop/namespaced/glue-workflow.yaml
@@ -4,6 +4,8 @@ stack_name: "{{ stack_group_config.namespace }}-glue-workflow"
 dependencies:
   - develop/namespaced/glue-tables.yaml
   - develop/namespaced/glue-job-S3ToJsonS3.yaml
+  - develop/namespaced/glue-job-JSONToParquet.yaml
+  - develop/namespaced/glue-job-compare-parquet.yaml
 parameters:
   Namespace: {{ stack_group_config.namespace }}
   JsonBucketName: {{ stack_group_config.intermediate_bucket_name }}
--- a/templates/glue-job-compare-parquet.yaml
+++ b/templates/glue-job-compare-parquet.yaml
@@ -1,0 +1,86 @@
+AWSTemplateFormatVersion: '2010-09-09'
+
+Description: The glue job for comparing parquet datasets
+
+Parameters:
+
+  JobDescription:
+    Type: String
+    Description: A fuller description of what the job does.
+    Default: ''
+
+  JobRole:
+    Type: String
+    Description: The name or ARN of the IAM role that will run this job.
+
+  S3ScriptBucket:
+    Type: String
+    Description: The name of the S3 bucket where the script file is located.
+
+  S3ScriptKey:
+    Type: String
+    Description: The bucket key where the script file is located.
+
+  MaxCapacity:
+    Type: Number
+    Description: How many DPUs to allot to this job.
+    Default: 1
+
+  MaxConcurrentRuns:
+    Type: Number
+    Description: >-
+      Number of this type of glue job that can be run at same time (double).
+    Default: 150
+
+  MaxRetries:
+    Type: Number
+    Description: How many times to retry the job if it fails (integer).
+    Default: 0 # TODO change this to 1 after initial development
+
+  GluePythonShellVersion:
+    Type: String
+    Description: The version of Python to use for the Glue Python Shell job
+
+  GlueVersion:
+    Type: String
+    Description: The version of glue to use for this job
+
+  TimeoutInMinutes:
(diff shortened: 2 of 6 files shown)


templates/glue-job-compare-parquet.yaml
```diff
@@ -0,0 +1,86 @@
+AWSTemplateFormatVersion: '2010-09-09'
```
rxu17, line 1: This cf template ended up being an exact copy of `glue-job-S3ToJsonS3.yaml`

templates/glue-workflow.j2
```diff
@@ -89,6 +103,29 @@ Resources:
       StartOnCreation: true
       WorkflowName: !Ref PrimaryWorkflow
 
+  JsontoParquetCompleteTrigger:
```
philerooski, line 106: We want this resource as well as the triggered glue job to be conditional on whether the namespace is `staging`.

templates/glue-job-compare-parquet.yaml
thomasyu888: +1 to Phil

config/develop/namespaced/glue-workflow.yaml
```diff
@@ -4,12 +4,17 @@ stack_name: "{{ stack_group_config.namespace }}-glue-workflow"
 dependencies:
   - develop/namespaced/glue-tables.yaml
   - develop/namespaced/glue-job-S3ToJsonS3.yaml
+  - develop/namespaced/glue-job-JSONToParquet.yaml
+  - develop/namespaced/glue-job-compare-parquet.yaml
```
thomasyu888, line 8: Just to confirm that it won't be an issue if we rename this and the S3ToJsonS3 to `glue-job-python-shell.yaml` and it just reduces the complexity.

thomasyu888 (commented):
🦖 Great work, I agree with Phil's comments and defer to him to approve after those are addressed.  Thanks!

templates/glue-workflow.j2
rxu17: The latter, and thanks for pointing this out. I almost forgot about this. :')

config/develop/namespaced/glue-workflow.yaml
rxu17: We are renaming the template: `/recover/templates/glue-job-S3ToJsonS3.yaml` to `glue-job-python-shell.yaml` but we will still have the separate stack configs (in this example):
- develop/namespaced/glue-job-JSONToParquet.yaml
- develop/namespaced/glue-job-compare-parquet.yaml

And they will update the [template path they are pointing to here](https://github.com/Sage-Bionetworks/recover/blob/main/config/develop/namespaced/glue-job-S3ToJsonS3.yaml#L2) to the renamed `glue-job-python-shell.yaml` under templates. Hope that alleviates your concern!

philerooski: That's perfect.

templates/glue-workflow.j2
rxu17: @philerooski 

From an infrastructure perspective, because I made our compare-parquet glue job into a separate resource, it will get created for any namespace no matter what just not triggered when I add in the condition to this trigger, however is it alright if the resource gets created when deploying to main. It would make it confusing when we look at the glue workflow console for a `main` workflow and see a non-active job...

Otherwise, I would create something like two conditions in the generic `glue-job-python-shell.yaml` that's like `IsNamespaceDependent` and `IsStagingNamespace` so we can only create the compare parquet dataset on specific conditions. But yeah this would involve updating the s3 to json stack configs to fullfil that. 

Or a 3rd way that I haven't considered? Maybe adding the compare parquet to the glue workflow template but that means a ton of parameters and code gets added there...

Do we have a preferred method?

thomasyu888: Agree with your Edit @rxu17

rxu17: **Update:** After some fiddling around, the 1st way seems the least complex mainly because if we conditionally create the compare parquet glue job, because it's a separate resource, there's just a lot of dependency errors such as the compare parquet job name that is being used by the glue workflow stack that it can't bypass.

On the bright side, with the 1st way, what happens is the compare parquet job does get created but it doesn't show up in the glue workflow console at all when namespace is main because it's not being triggered by anything and will only run on a "feature" namespace. 

That is the resolution i have found for now.

thomasyu888 (approved):
🔥 LGTM!

rxu17:
Just tried testing the glue shell job and will have to add modifications to the CF templates like having it include `datacompy` as an external python module.

rxu17:
This PR is ready for final review. I tested this by copying over the compare parquet datasets script from etl-409, and if you look under the etl-406-glue-workflow runs, you can see the successful completion of the workflow. 

You can also look in the relevant bucket to find the comparison outputs.

templates/glue-workflow.j2
philerooski: We use the `!Ref AWS::NoValue` method in `templates/s3-bucket.yaml`

templates/glue-job-python-shell.yaml
```diff
@@ -83,6 +101,6 @@ Resources:
 Outputs:
 
   JobName:
-    Value: !Ref GlueJob
+    Value: !If [CreateResource, !Ref GlueJob, "GlueJobPythonShellNotCreated"]
```
philerooski, line 104: Use `!Ref 'AWS::NoValue'` in the case we do not create the resource

rxu17: Yeah I tried that originally. Apparently in Outputs, they don't allow a `!Ref 'AWS::NoValue'`, it has to be a string. 

Here was the full message: `"An error occurred (ValidationError) when calling the UpdateStack operation: Template format error: The Value field of every Outputs member must evaluate to a String."`

templates/glue-workflow.j2
rxu17: Right, I see it under `BucketName` under `Properties` for the Bucket resource. I think it still creates the Bucket resource, just doesn't output a BucketName.

philerooski (approved):

PR #44, opened 2023-04-27
[ETL-414] Add script to generate test S3 event notification

philerooski:
This script can generate a S3 event notification test event for a single S3 object or for all S3 objects which exist directly under a specific S3 key prefix. Using this test event with our lambda simulates an object create/update event in an S3 bucket.

### DIFF
--- a/src/lambda_function/README.md
+++ b/src/lambda_function/README.md
@@ -35,20 +35,17 @@ sam build
 
 ### Creating/modifying test events
 
-The `test-trigger-event.json` at `src/lambda_function/events` contains a
-dummy event of what a scheduled event trigger using CloudWatch would look
-like as our lambda function receives it. The part of the event that is actually used
-is the `time` key in the file whose date portion is used in the lambda function to
-compare against files in the S3 bucket we are checking against since only files that were
-modified/created the same date as the trigger event would be submitted by the lambda to
-the s3-to-glue workflow. You can modify the value for the `time` to test whether files
-get submitted or not.
+The file `single-record.json` in `src/lambda_function/events` contains a
+dummy event for an S3 event trigger. You can generate your own test events
+for single or multiple records with the script at
+`src/lambda_function/events/generate_test_event.py`.
 
 ### Invoking test events
 
 To test the lambda function locally, run the following command from the lambda directory.
-Use `test-trigger-event.json` to trigger the Glue workflow. The file `test-env-vars.json` contains
-the namespace environment variable that is expected by the lambda script.
+Use `single-record.json` or your own test event to trigger the S3 to JSON Glue workflow.
+The file `test-env-vars.json` contains
+the environment variable that is expected by the lambda script.
 Don't forget to update the value of this variable
 if you are testing a stack deployed as part of a feature branch.
 
@@ -56,7 +53,7 @@ To invoke the lambda with the test event:
 
 ```bash
 cd src/lambda_function
-sam local invoke -e events/test-trigger-event.json --env-vars test-env-vars.json
+sam local invoke -e events/single-record.json --env-vars test-env-vars.json
 ```
 
 ## Launching Lambda stack in AWS
--- a/src/lambda_function/events/generate_test_event.py
+++ b/src/lambda_function/events/generate_test_event.py
@@ -1,66 +1,46 @@
 """
 This is a utility script that generates a fake S3 event notification and writes
-it out as JSON. This test event can be used to test the lambda which
+it out as JSON. This test event can be used to test the Lambda which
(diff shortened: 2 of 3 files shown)


src/lambda_function/events/generate_test_event.py
```diff
@@ -0,0 +1,138 @@
...
+    parser = argparse.ArgumentParser(
+      description="Generate a json file of a mocked SQS event for testing.")
+    parser.add_argument("--input-bucket",
+      default="recover-dev-input-data",
+      help="Optional. S3 bucket name containing input data. Default is 'recover-dev-input-data'.")
+    parser.add_argument("--input-key",
+      default="main/2023-01-12T22--02--17Z_77fefff8-b0e2-4c1b-b0c5-405554c92368",
+      help=("Optional. Name of the Synapse dataset containing test data. "
+            "Default is `main/2023-01-12T22--02--17Z_77fefff8-b0e2-4c1b-b0c5-405554c92368` "))
+    parser.add_argument("--input-key-prefix",
```
rxu17, line 60: I just realize for recover data, if we reprocess, we can only reprocess everything at this point or if we somehow know the specific filename? Are the individual recover zipped files grouped by anything in particular other than the datetime it was created?

rxu17 (approved):
Looks good to me!

src/lambda_function/events/generate_test_event.py
philerooski: > Are the individual recover zipped files grouped by anything

That's a discussion we could have with digital health. They are the ones copying the data from the ingestion bucket to the input bucket, so we could come up with a partitioning structure.

In practice we would only ever submit the entire dataset. We need to remove deleted samples and those could appear anywhere.

src/lambda_function/events/generate_test_event.py
```diff
@@ -0,0 +1,138 @@
...
+
+import argparse
+import copy
+import json
+import boto3
+
+SINGLE_RECORD_OUTFILE = 'single-record.json'
+MULTI_RECORD_OUTFILE = 'records.json'
+
+s3_event_record_template = {
```
thomasyu888, line 15: Nit: this is a global variable so it should be all caps.

src/lambda_function/events/generate_test_event.py
```diff
@@ -0,0 +1,138 @@
...
+    args = read_args()
+    print("Generating mock S3 event...")
+    s3_event = create_event(
+            bucket=args.input_bucket,
+            key=args.input_key,
+            key_prefix=args.input_key_prefix
+    )
+
+    if args.input_key_prefix is not None:
+        with open(MULTI_RECORD_OUTFILE, "w") as outfile:
```
thomasyu888, line 127: Since we don't expect this script to live on forever, its probably fine to not have these as parameters.

src/lambda_function/events/generate_test_event.py
```diff
@@ -0,0 +1,138 @@
+"""
+This is a utility script that generates a fake S3 event notification and writes
+it out as JSON. This test event can be used to test the lambda which
```
thomasyu888, line 3: Can you speak to how the json can be used to test the lambda?

thomasyu888 (commented):
Awesome - just some questions/comments prior to approval.

src/lambda_function/events/generate_test_event.py
philerooski: I expanded the docstring a bit and left a link to the README (which I also updated).

src/lambda_function/events/generate_test_event.py
```diff
@@ -0,0 +1,144 @@
...
+import json
+import boto3
+
+SINGLE_RECORD_OUTFILE = 'single-record.json'
+MULTI_RECORD_OUTFILE = 'records.json'
+
+
+def read_args() -> argparse.Namespace:
+    parser = argparse.ArgumentParser(
+      description="Generate a json file of a mocked SQS event for testing.")
```
thomasyu888, line 25: Nit: Try this, then you can remove your "Default" strings
```suggestion
    parser = argparse.ArgumentParser(
      description="Generate a json file of a mocked SQS event for testing.",
      formatter_class=argparse.ArgumentDefaultsHelpFormatter
    )
```

src/lambda_function/README.md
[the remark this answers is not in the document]
philerooski: No. Are you wondering how we will submit these events as part of our tests or as part of the production pipeline?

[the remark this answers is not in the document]
thomasyu888: Thanks @philerooski. ^ Both

PR #45, opened 2023-05-02
[ETL-434] Modify S3 to Glue Lambda code to use S3 event notifications

philerooski:
Effectively a rewrite of our S3 to Glue Lambda to use S3 event notifications. We skirt the owner.txt issues identified in the JIRA ticket by ignoring any objects which contain "owner.txt" in their name.

### DIFF
--- a/src/lambda_function/s3_to_glue/app.py
+++ b/src/lambda_function/s3_to_glue/app.py
@@ -1,74 +1,32 @@
 import os
 import json
 import logging
-import datetime
-from dateutil import parser
-
 import boto3
 
 logger = logging.getLogger()
 logger.setLevel(logging.INFO)
 
-
-def query_files_to_submit(
-    objects: dict, files_to_submit: list, trigger_event_date: datetime
-):
-    """Queries the files in bucket for ones with the same
-    date as the trigger event
-
-    Args:
-        objects (dict): bucket objects' metadata info
-        files_to_submit (list): fielsto submit from source s3 bucket
-        trigger_event_date (datetime): _description_
-
-    Returns:
-        list: List of dictionaries containing source_bucket and source_key
-        for each file to submit from the source s3 bucket
+def submit_s3_to_json_workflow(messages: list[dict[str,str]], workflow_name):
     """
-    files_current = objects["Contents"]
-    logger.info("Querying S3 bucket for new data from source")
-
-    # query bucket for new/updated files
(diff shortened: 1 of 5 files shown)


src/lambda_function/s3_to_glue/app.py
```diff
@@ -86,20 +50,19 @@ def lambda_handler(event, context):
...
-    trigger_event_date = parser.isoparse(event["time"]).date()
-    objects = s3.list_objects_v2(Bucket=source_bucket_name)
-    files_to_submit = []
-    if "Contents" in objects:
-        files_to_submit = query_files_to_submit(
-            objects, files_to_submit, trigger_event_date
+    messages = []
+    for record in event["Records"]:
+        bucket_name = record["s3"]["bucket"]["name"]
+        object_key = record["s3"]["object"]["key"]
```
thomasyu888, line 56: Just a note here: https://docs.aws.amazon.com/AmazonS3/latest/userguide/notification-content-structure.html.  It was helpful to know the S3 event notification structure.

One thing I'm not fully understanding is that this seems to just take in all the new records within a bucket without specific folders.  Based on what you were saying, am I understanding correctly that we would set up S3 notifications specific to different folders?

thomasyu888 (commented):
Nice! just a question prior to approval

src/lambda_function/s3_to_glue/app.py
philerooski: The trigger can be configured to only alert for objects under a specific prefix, e.g., `{namespace}/raw_data/...`

rxu17 (approved):
Lgtm!

tests/test_s3_to_glue_lambda.py
```diff
@@ -1,108 +1,70 @@
...
+        return messages
 
-def test_empty_bucket_contents(test_trigger_event):
-    test_submit_files = app.query_files_to_submit(
-        objects={"Contents": []},
-        files_to_submit=[],
-        trigger_event_date=parser.isoparse(test_trigger_event["time"]).date(),
-    )
-    assert test_submit_files == []
+    def test_submit_s3_to_json_workflow(self, test_messages, monkeypatch):
```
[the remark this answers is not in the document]
philerooski: I wanted to use moto but moto does not support most methods of the Glue client.

PR #46, opened 2023-05-04
[ETL-374] Fix failing test for setup_external_storage

rxu17:
**Purpose:** This fixes the failing integration test for `setup_external_storage.py` and allows for the addition of command line arguments for running the test so that you are able to test STS access for multiple synapse folders that had STS access setup rather than a hard coded folder.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -56,10 +56,23 @@ pytest with other tests because they have to be run in a Dockerfile:
 - test_s3_to_glue_lambda.py
 - test_setup_external_storage.py
 
-Example)
+
+#### Running tests for lambda
 Run the following command from the repo root to run tests for the lambda function (in develop).
-You can run this locally or inside the docker image.
 
 ```shell script
 python3 -m pytest tests/test_s3_to_glue_lambda.py -v
+```
+
+#### Running tests for setup external storage
+Run the following command from the repo root to run the integration test for the setup external storage script to check that the STS
+access has been set for a given synapse folder (in develop).
+
+This test takes in two command line arguments:
+
+- test-synapse-folder-id - synapse id of the folder to check STS access for
+- test-ssm-parameter - ssm parameter to get AWS credentials for otherwise leave blank and it will pull credentials from the environment
+
+```shell script
+python3 -m pytest tests/test_setup_external_storage.py --test-synapse-folder-id <put_synapse_folder_id_here> --test-ssm-parameter <put_ssm_parameter_here_or_leave_blank>
 ```
--- a/tests/test_setup_external_storage.py
+++ b/tests/test_setup_external_storage.py
@@ -8,12 +8,12 @@ from src.scripts.setup_external_storage import setup_external_storage
 
 @pytest.fixture()
 def test_synapse_folder_id(pytestconfig):
-    return pytestconfig.getoption("test_synapse_folder_id")
+    yield pytestconfig.getoption("test_synapse_folder_id")
 
 
 @pytest.fixture()
 def test_ssm_parameter(pytestconfig):
-    return pytestconfig.getoption("test_ssm_parameter")
+    yield pytestconfig.getoption("test_ssm_parameter")
 
 
 def test_setup_external_storage_success(test_synapse_folder_id, test_ssm_parameter):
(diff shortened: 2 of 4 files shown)


thomasyu888 (approved):
🔥 LGTM!

tests/README.md
```diff
@@ -56,10 +56,23 @@ pytest with other tests because they have to be run in a Dockerfile:
...
 
 ```shell script
 python3 -m pytest tests/test_s3_to_glue_lambda.py -v
 ```
+
+#### Running tests for setup external storage
+Run the following command from the repo root to run the integration test for the setup external storage script to check that the STS
+access has been set for a given synapse folder (in develop).
+
+This test takes in two command line arguments:
```
philerooski, line 71: When running the entire test suite, do we also need to provide these two command line arguments?

tests/README.md
```diff
@@ -56,10 +56,23 @@ pytest with other tests because they have to be run in a Dockerfile:
...
+
+- test-synapse-folder-id - synapse id of the folder to check STS access for
```
philerooski, line 73: I think this is a good opportunity to use the Bridge Downstream Test project: https://www.synapse.org/#!Synapse:syn26721259/wiki/

This is a permanent Synapse project (i.e., it's not created/destroyed programmatically) that we can use to test Synapse functionality that's too complicated to mock (we also use it to maintain test datasets). Rather than requiring a tester to create their own Synapse project and/or folder before running tests, programmatically create/destroy a test folder in this project as part of the test setup/teardown

tests/test_setup_external_storage.py
```diff
@@ -1,52 +1,38 @@
...
-    parquet files in the S3 bucket location to verify that it has access"""
+    files in the S3 bucket location to verify that it has access"""
+    test_synapse_client = setup_external_storage.get_synapse_client(
+        ssm_parameter=test_ssm_parameter, aws_session=boto3
+    )
     # Get STS credentials
     token = test_synapse_client.get_sts_storage_token(
-        entity=test_parquet_folder, permission="read_only", output_format="json"
+        entity=test_synapse_folder_id, permission="read_only", output_format="json"
     )
```
philerooski, line 28: What happens if we try to get an STS token for a storage location which is not properly configured for use with Synapse? My guess is that Synapse will throw an exception and everything below this line is doing more to test the `pyarrow` package than it is to test our `setup_external_storage` script.

rxu17: Yes, we could leave out the lines after this, because this is the line that essentially tests the STS access

tests/README.md
rxu17: Yes, if we wanted to run all of the tests. However, our section regarding general running all of the tests needs to be updated because we currently can't run all of our tests under the same docker image nor all of our tests locally. They need separate docker images, which is why I didn't modify this section so this should be handled when we implement the tests into our CI/CD.

tests/README.md
rxu17: I was hoping this test could be used in more of an integration test way and be used to continuously test (if not continuously test, then check) that there is STS access to the expected synapse projects for recover data each time we deploy or something. Thoughts on that?

rxu17: Sounds good! Made the changes

philerooski: Thanks! But won't you need to do setup/teardown of the test folder? If you use a persistent folder then you may successfully configure the storage location during one test run and not notice breaking changes during the next test run. And the folder you used (syn27558289) is already being used for something else. We don't want to be modifying its external storage location.

philerooski: Also I'm just realizing this is a RECOVER script so it's a bit silly to suggest we use the Bridge Downstream test project for it 😕

rxu17: Hmmm, but we're not configuring the storage location on the test folder?

I was thinking of this test script used in a way like every time we deploy, there's a GH action that runs this test on the synapse folders that are part of the dev pipeline to check that they have the expected STS access.

Similarly, same thing can happen with the staging pipeline in prod.

Because we run `setup_external_storage.py` manually whenever we need to setup STS access so this part where you mention 

> If you use a persistent folder then you may successfully configure the storage location during one test run and not notice breaking changes during the next test run. And the folder you used (syn27558289) is already being used for something else. We don't want to be modifying its external storage location.

won't be a concern?

philerooski: Ohh I see. I have been thinking of everything under `tests/` as quasi-unit tests ("quasi" because sometimes we don't mock things and that puts some of these tests in integration test territory). But although this file is called `test_setup_external_storage`, it's not actually testing our `setup_external_storage` script.

> I was thinking of this test script used in a way like every time we deploy, there's a GH action that runs this test on the synapse folders that are part of the dev pipeline to check that they have the expected STS access.

This is more of an integration test. It's fine to leave it as it is if that's your intention.

rxu17: Yup, I'm not sure if integration tests handled by pytest also belong also under `tests/` (I've seen that being the case before). 

I will leave it as is (will remove the default synapse folder though)

PR #50, opened 2023-05-10
[ETL-397] Create s3-event-config lambda and dependencies to add s3 notification configuration

rxu17:
**Purpose:** This PR adds the following resources seen here: [ETL-397](https://sagebionetworks.jira.com/browse/ETL-397). 

This allows us to add the S3 notification configuration to our existing input data buckets (via the new lambda), and also allow our S3 to glue lambda to now be triggered by those S3 new object notifications and start our recover ETL processing. See the ticket for why we were not able to add those changes to our pre-existing lambda resources/stacks/templates and had to do it through a separate lambda function.

I put this PR up as a draft in case there's some refactoring/structural changes we would like to make. Still testing it on the pipeline but it works for one dataset I uploaded. Also I separated out the implementation of the GH action to trigger this lambda here in [ETL-446](https://sagebionetworks.jira.com/browse/ETL-446)

[ETL-397]: https://sagebionetworks.jira.com/browse/ETL-397?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ETL-446]: https://sagebionetworks.jira.com/browse/ETL-446?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/lambda_function/s3_event_config/app.py
thomasyu888, line 1: add an `__init__.py`

src/lambda_function/s3_event_config/app.py
```diff
@@ -0,0 +1,67 @@
...
+            logger.info("Sending response to custom resource after Delete")
+
+        elif event["RequestType"] == "Create" or event["RequestType"] == "Update":
+            logger.info(f'Request Type: {event["RequestType"]}')
+            lambda_arn = os.environ["S3_TO_GLUE_FUNCTION_ARN"]
+            bucket = os.environ["S3_SOURCE_BUCKET_NAME"]
+            add_notification(lambda_arn, bucket)
+            logger.info("Sending response to custom resource")
+    except Exception as e:
+        logger.info(f"Failed to process:{e}")
```
thomasyu888, line 38: I'm assuming these will be written as cloudwatch logs?

src/lambda_function/s3_event_config/app.py
```diff
@@ -0,0 +1,67 @@
...
+import boto3
+
+logger = logging.getLogger()
+logger.setLevel(logging.INFO)
+
+SUCCESS = "SUCCESS"
+FAILED = "FAILED"
+
+logger.info("Loading function")
+s3 = boto3.resource("s3")
```
thomasyu888, line 19: Nit: I don't have strong feelings about this, but what are your thoughts about including this within the functions themselves?

src/lambda_function/s3_event_config/app.py
```diff
@@ -0,0 +1,67 @@
+"""
+This Lambda app responds to an S3 event notification and starts a Glue workflow.
```
thomasyu888, line 2: Does this description need to be updated?

src/lambda_function/s3_event_config/app.py
```diff
@@ -0,0 +1,67 @@
...
+FAILED = "FAILED"
+
+logger.info("Loading function")
+s3 = boto3.resource("s3")
+
+
+def lambda_handler(event, context):
+    logger.info(f"Received event: {json.dumps(event, indent=2)}")
+    try:
+        if event["RequestType"] == "Delete":
```
thomasyu888, line 25: Could you point me to the list of supported `RequestType`?

src/lambda_function/s3_event_config/app.py
```diff
@@ -0,0 +1,67 @@
...
+        NotificationConfiguration={
+            "LambdaFunctionConfigurations": [
+                {"LambdaFunctionArn": lambda_arn, "Events": ["s3:ObjectCreated:*"]}
+            ]
+        }
+    )
+    logger.info("Put request completed....")
+
+
+def delete_notification(bucket: str) -> None:
```
thomasyu888, line 59: Just want to make sure that I understand that this is a lambda that adds the lambda function that will trigger jobs for S3 event notifications.  

If so, how will the delete notification be triggered?

thomasyu888 (commented):
Great start, some initial comments.

src/lambda_function/s3_event_config/app.py
rxu17: Yes they show up in the cloudwatch logs

src/lambda_function/s3_event_config/app.py
rxu17: This is defined by us. So we can change this to be whatever we'd like if that makes sense.

src/lambda_function/s3_event_config/app.py
rxu17: image
A rough diagram of where this all falls in place. The delete notification is in place in case we'd like to delete the notification configuration/would like a more automated way of deleting the configuration. If we'd rather do that via AWS console, we can remove that code.

src/lambda_function/s3_event_config/app.py
rxu17: I was planning to add them to the README

src/lambda_function/s3_event_config/app.py
rxu17: Actually, I might even remove the try, catch so it will just error out instead of showing that it was "successful".

philerooski (changes requested):
Amazing 🤯 

I'm curious, how does this play out in staging vs main namespace in the production account? When we merge a PR does the event trigger get configured in staging? This would cause any new data that comes in to be processed by staging as well, which I don't think is desirable.

BUT, maybe that's not necessarily the case. If we don't trigger staging's `s3_event_config` lambda in the Github CI/CD then the S3 event trigger never gets configured for staging, right?

rxu17:
> Amazing 🤯
> 
> I'm curious, how does this play out in staging vs main namespace in the production account? When we merge a PR does the event trigger get configured in staging? This would cause any new data that comes in to be processed by staging as well, which I don't think is desirable.
> 
> BUT, maybe that's not necessarily the case. If we don't trigger staging's `s3_event_config` lambda in the Github CI/CD then the S3 event trigger never gets configured for staging, right?

Yes, as long as we don't trigger staging's `s3_event_config` lambda in the Github CI/CD, the S3 event trigger should not get triggered for any data that gets put into the staging namespace folder of the input data bucket.

rxu17:
From my research (see ticket for more info), there doesn't exist a way to allow the S3 to JSON lambda to process multiple S3 object notifications as a group besides implementing the SQS architecture. I've created another ticket [ETL-450](https://sagebionetworks.jira.com/browse/ETL-450) to explore this.

rxu17:
Also, I tested this on copying one dataset (using the recover pilot datasets) at a time to the bucket (and letting the glue workflow run all the way through), and I wasn't able to reproduce the random JSON to Parquet job errors I got previously. I feel like that's likely due to one or more of the following:

- I didn't upload/copy more than one file to the bucket at a time and waited for the glue workflow to run completely
- I only had S3 notification happen when a file gets uploaded to the specific namespace of the input data bucket
- Glue was being weird

Pipfile
```diff
@@ -16,3 +16,4 @@ synapseclient = "~=2.7"
 pandas = "<1.5"
 moto = "~=4.1"
 datacompy = "~=0.8"
+docker = "~=6.1"
```
philerooski, line 19: Why do we need this?

philerooski (approved):

Pipfile
rxu17: It's dependency of using `mock_lambda` from `moto3` but for some reason installing `moto3` doesn't automatically install `docker` as a dependency

thomasyu888 (approved):
🔥 LGTM!

PR #53, opened 2023-05-22
Include digital health service account in owner.txt

philerooski:
Allows the service account to use the pre-ETL bucket in both develop and production as a Synapse external storage location.


thomasyu888 (approved):
🔥 LGTM!

PR #55, opened 2023-06-01
[ETL-446] Add upload-and-deploy-to-prod-main workflow

philerooski:
This ticket was for work needed to configure the S3 event trigger on the S3 to Glue lambda. Since this only occurs in prod/main, I've written a separate github workflow which deploys to prod/main and configures this trigger. This workflow is run whenever we push a semantically versioned tag to this repo.

### DIFF
--- a/.github/workflows/upload-and-deploy-to-prod-main.yaml
+++ b/.github/workflows/upload-and-deploy-to-prod-main.yaml
@@ -71,7 +71,7 @@ jobs:
         run: pipenv run sceptre --var "namespace=${{ env.NAMESPACE }}" launch prod --yes
 
       - name: Configure S3 to Glue lambda with S3 trigger
-        uses: actions/invoke-aws-lambda@v3
+        uses: gagoar/invoke-aws-lambda@v3
         with:
           AWS_ACCESS_KEY_ID: ${{ env.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ env.AWS_SECRET_ACCESS_KEY }}
--- a/.github/workflows/upload-and-deploy-to-prod-main.yaml
+++ b/.github/workflows/upload-and-deploy-to-prod-main.yaml
@@ -1,0 +1,82 @@
+name: upload-and-deploy-to-prod-main
+
+description: >-
+  Upload artifacts and deploy stacks to main when a new tag is pushed.
+  Tags must be formatted using semantic versioning.
+
+on:
+  push:
+    tags:
+      - "v[0-9]+.[0-9]+.[0-9]+"
+
+env:
+  NAMESPACE: main
+  PYTHON_VERSION: 3.9
+
+jobs:
+
+  pre-commit:
+    name: Run pre-commit hooks against all files
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+      - uses: pre-commit/action@v3.0.0
+
+
+  upload-files:
+    name: Upload files to S3 bucket in prod/main
+    runs-on: ubuntu-latest
+    needs: pre-commit
+    # These permissions are needed to interact with GitHub's OIDC Token endpoint.
+    permissions:
+      id-token: write
+      contents: read
+    environment: prod
+    steps:
+
+      - name: Setup code, pipenv, aws
+        uses: Sage-Bionetworks/action-pipenv-aws-setup@v3
+        with:
+          role_to_assume: ${{ vars.AWS_CREDENTIALS_IAM_ROLE }}
+          role_session_name: GitHubActions-${{ github.repository_owner }}-${{ github.event.repository.name }}-${{ github.run_id }}
+          python_version: ${{ env.PYTHON_VERSION }}
+
+      - name: Copy files to templates bucket
+        run: python src/scripts/manage_artifacts/artifacts.py --upload --namespace $NAMESPACE --cfn_bucket ${{ vars.CFN_BUCKET }}
+
+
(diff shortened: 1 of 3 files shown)


.github/workflows/upload-and-deploy-to-prod-main.yaml
```diff
@@ -0,0 +1,82 @@
...
+      - name: Deploy sceptre stacks to prod/main
+        run: pipenv run sceptre --var "namespace=${{ env.NAMESPACE }}" launch prod --yes
+
+      - name: Configure S3 to Glue lambda with S3 trigger
+        uses: actions/invoke-aws-lambda@v3
+        with:
+          AWS_ACCESS_KEY_ID: ${{ env.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ env.AWS_SECRET_ACCESS_KEY }}
+          AWS_SESSION_TOKEN: ${{ env.AWS_SESSION_TOKEN }}
+          REGION: ${{ env.AWS_REGION }}
```
philerooski, line 79: These environment variables are exported by aws-actions/configure-aws-credentials, which we use within Sage-Bionetworks/action-pipenv-aws-setup.

src/lambda_function/s3_event_config/template.yaml
```diff
@@ -45,6 +46,13 @@ Resources:
           S3_TO_GLUE_FUNCTION_ARN: !Ref S3ToGlueFunctionArn
           BUCKET_KEY_PREFIX: !Ref Namespace
 
+  LambdaInvokePermission:
+    Type: AWS::Lambda::Permission
+    Properties:
+      FunctionName: !GetAtt S3ToGlueFunction.Arn
+      Action: lambda:InvokeFunction
+      Principal: !Ref 'AWS::AccountId'
+
```
philerooski, line 55: This invoke permission is relatively broad, but [the docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-permission.html) don't mention being able to specify a specific IAM role. I couldn't find any additional information on this use case.

.github/workflows/upload-and-deploy-to-prod-main.yaml
```diff
@@ -0,0 +1,82 @@
...
+          role_session_name: GitHubActions-${{ github.repository_owner }}-${{ github.event.repository.name }}-${{ github.run_id }}
+          python_version: ${{ env.PYTHON_VERSION }}
+
+      - name: Create directory for remote sceptre templates
+        run: mkdir -p templates/remote/
+
+      - name: Deploy sceptre stacks to prod/main
+        run: pipenv run sceptre --var "namespace=${{ env.NAMESPACE }}" launch prod --yes
+
+      - name: Configure S3 to Glue lambda with S3 trigger
```
rxu17, line 73: Was there a reason to leave this lambda trigger action out of the `upload-and-deploy` for dev?

src/lambda_function/s3_event_config/template.yaml
```diff
@@ -45,6 +46,13 @@ Resources:
           S3_TO_GLUE_FUNCTION_ARN: !Ref S3ToGlueFunctionArn
           BUCKET_KEY_PREFIX: !Ref Namespace
 
+  LambdaInvokePermission:
+    Type: AWS::Lambda::Permission
+    Properties:
+      FunctionName: !Ref S3ToGlueFunctionArn
```
rxu17, line 52: Should this be invoking the Arn of `S3EventConfigFunction`?

.github/workflows/upload-and-deploy-to-prod-main.yaml
philerooski: I wasn't sure if it would be useful. It could be a nuisance if every time we move data into a namespace in the input bucket the lambda is triggered. Should I be doing this in dev, too?

src/lambda_function/s3_event_config/template.yaml
philerooski: Yes.

.github/workflows/upload-and-deploy-to-prod-main.yaml
rxu17: True, I think the main reason I was thinking of having it in dev as well to mirror the structure is if the `lambda action` or something else related to the lambda trigger broke when we are deploying something in dev, we could catch it then (plus one less difference to keep track of between dev and prod). I'm not sure if that's enough of a reason to add it to dev?

src/lambda_function/s3_event_config/template.yaml
philerooski: It's also unclear to me if this is even necessary. The IAM role assumed by the github action has admin permissions and we might not need an additional `AWS::Lambda::Permission` here. I will do some testing.

.github/workflows/upload-and-deploy-to-prod-main.yaml
philerooski: Yep, I added it to dev.

src/lambda_function/s3_event_config/template.yaml
philerooski: We don't need it.

thomasyu888 (approved):
🔥 Nice CI/CD here.  We may want to look into not deploying the semantic versioned tags (for the other github action) since the workflow fails with the namespace name of "v0.2.0"

philerooski:
> 🤖 Nice CI/CD here. We may want to look into not deploying the semantic versioned tags (for the other github action) since the workflow fails with the namespace name of "v0.2.0"

Nice catch! I updated the `on` field to ignore tags.

PR #57, opened 2023-06-08
[ETL-466] Archive existing parquet datasets before writing parquet

philerooski:
Archives existing datasets in S3 by moving them to a timestamped subfolder within an "archive" folder. The format of the timestamped subfolder is:
```
{year}_{month}_{day}_{workflow_run_id}
```
If the dataset does not exist at `key_prefix` or if the dataset is empty, no action will be taken.

Also added some logging statements to `add_index_to_table`.

### DIFF
--- a/src/glue/jobs/json_to_parquet.py
+++ b/src/glue/jobs/json_to_parquet.py
@@ -13,8 +13,9 @@ assessmentid, year, month, and day to each record in each table.
 
 import os
 import sys
-import boto3
+from datetime import datetime
 
+import boto3
 from awsglue import DynamicFrame
 from awsglue.context import GlueContext
 from awsglue.job import Job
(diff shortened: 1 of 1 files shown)


philerooski:
@thomasyu888 Are we waiting on feedback from digital health before pulling this in? I don't see a need to get anything beyond an acknowledgment of the design from them, but right now we don't even have that so not sure if I should convert this to a regular PR.

thomasyu888:
@philerooski can we get an acknowledgement from them?

src/glue/jobs/json_to_parquet.py
```diff
@@ -182,11 +183,86 @@ def drop_deleted_healthkit_data(
...
+    Raises:
+        N/A
+    """
+    logger = glue_context.get_logger()
+    s3_client = boto3.client("s3")
+    response = s3_client.list_objects_v2(
+            Bucket=bucket,
+            Prefix=key_prefix
+    )
+    if "Contents" not in response:
```
thomasyu888, line 223: In what scenario will there not be "Contents"?

src/glue/jobs/json_to_parquet.py
```diff
@@ -182,11 +183,86 @@ def drop_deleted_healthkit_data(
...
+    formatted_datetime = ( # e.g., 2023_06_28
+            str(datetime.now())
+            .split(" ")[0]
+            .replace('-', '_')
+    )
+    this_archive_collection = "_".join([formatted_datetime, workflow_run_id])
+    source_objects = [
+            obj for obj in response["Contents"] if obj["Size"] > 0]
+    objects_to_delete = []
+    for source_object in source_objects:
```
thomasyu888, line 234: Nit: This is trivial, but you are looping through this list twice.  

I think its fine - what is your rationale for doing this?

src/glue/jobs/json_to_parquet.py
```diff
@@ -182,11 +183,86 @@ def drop_deleted_healthkit_data(
...
+                CopySource=this_object,
+                Bucket=bucket,
+                Key=destination_key
+        )
+        this_object.pop("Bucket")
+        objects_to_delete.append(this_object)
+    if delete_upon_completion:
+        s3_client.delete_objects(
+                Bucket=bucket,
+                Delete={"Objects": objects_to_delete}
```
thomasyu888, line 257: Will this line of code fail if `object_to_delete` is empty?

src/glue/jobs/json_to_parquet.py
```diff
@@ -253,13 +338,16 @@ def add_index_to_table(
     Returns:
         awsglue.DynamicFrame with index columns
     """
+    logger = glue_context.get_logger()
     _, table_data_type = table_name.split("_")
     this_table = unprocessed_tables[table_key].toDF()
+    logger.info(f"Adding index to {table_name}")
     if table_key == table_name: # top-level fields already include index
         for c in list(this_table.columns):
```
thomasyu888, line 346: Nit: I'm not a huge fan of single letter variable because they aren't as descriptive, but here, I understand `c` is a `column`

thomasyu888 (commented):
Nice - please see some of my comments.

src/glue/jobs/json_to_parquet.py
```diff
@@ -182,11 +183,86 @@ def drop_deleted_healthkit_data(
...
+    """
+    logger = glue_context.get_logger()
+    s3_client = boto3.client("s3")
+    response = s3_client.list_objects_v2(
+            Bucket=bucket,
+            Prefix=key_prefix
+    )
+    if "Contents" not in response:
+        return
+    formatted_datetime = ( # e.g., 2023_06_28
```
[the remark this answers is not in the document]
philerooski: The "archive" folder name includes the workflow run ID. So each workflow run will archive to a unique folder.

src/glue/jobs/json_to_parquet.py
philerooski: If the key prefix does not exist or if it does exist but is empty. That would mean that this is the first time we are processing this data type (or all the data for this data type was deleted for some reason) and there is nothing to archive.

src/glue/jobs/json_to_parquet.py
philerooski: It's prettier than
```
for source_object in [obj for obj in response["Contents"] if obj["Size"] > 0]
```

src/glue/jobs/json_to_parquet.py
philerooski: `object_to_delete` might be empty if the data type key prefix exists but there are no objects in it? It does throw an error if `object_to_delete` is empty so I added the condition that it must not be empty.

thomasyu888 (approved):
🔥 excellent!  LGTM.

PR #58, opened 2023-06-13
[ETL-333] Add tests for JSON to Parquet Glue job

philerooski:
I adapted the tests we were using in Bridge Downstream for Recover and added some additional tests. The tests take the approach of not mocking complicated objects like GlueContext and DynamicFrame (which rely on a Spark backend), and instead creating actual resources in a testing namespace so that we can load these resources as DynamicFrames and see what affect each function has. I did try to mock the AWS resources, but this ultimately didn’t help because all of the actual functionality is happening in a Glue/Spark environment and the resources are only helpful in obtaining a DynamicFrame to operate on. Calling GlueContext.create_dynamic_frame.from_catalog and passing a mocked AWS database and table raises a Py4JJavaError, so if we were to try to use a pure unit-test-style approach we could ignore the AWS resources entirely (except perhaps S3) and try mocking GlueContext etc. instead. This was determined to be too time-consuming to complete within the current sprint, so I kept the “dirty” approach of using actual resources in the tests.


tests/test_json_to_parquet.py
```diff
@@ -0,0 +1,531 @@
...
+        )
+        if "Contents" in response:
+            objects_to_delete = []
+            for obj in response["Contents"]:
+                objects_to_delete.append({"Key": obj["Key"]})
+            if objects_to_delete:
+                s3_client.delete_objects(
+                    Bucket=artifact_bucket,
+                    Delete={"Objects": objects_to_delete}
+                )
```
thomasyu888, line 52: When you yield here, does the rest of this code execute?

tests/test_json_to_parquet.py
```diff
@@ -0,0 +1,531 @@
...
+
+@patch(
+        "src.glue.jobs.json_to_parquet.INDEX_FIELD_MAP",
+        {"TestFlatDataType": ["GlobalKey"], "TestNestedDataType": ["GlobalKey"]}
+)
+class TestJsonS3ToParquet:
+
+    @pytest.fixture(scope="class")
+    def glue_database_name(self, namespace):
+        return f"{namespace}-pytest-database"
```
thomasyu888, line 22: Nit: Can these be `yield` instead of `return`?

tests/test_json_to_parquet.py
```diff
@@ -0,0 +1,531 @@
...
+                        }
+                    },
+                    "Grouping": {
+                        "TableGroupingPolicy":"CombineCompatibleSchemas"
+                    }
+                })
+        )
+        glue_client.start_crawler(Name=crawler_name)
+        response = {"Crawler": {}}
+        for i in range(60): # wait up to 10 minutes for crawler to finish
```
thomasyu888, line 305: nit: should this be a `while` loop instead?

tests/test_json_to_parquet.py
```diff
@@ -0,0 +1,531 @@
...
+        )
+        assert (
+                set(tables_with_index[table_key].schema.fieldNames()) ==
+                set(['id', 'index', 'filename', 'timestamp', 'GlobalKey'])
+        )
+        child_table_with_index = (tables_with_index[table_key]
+              .toPandas()
+              .sort_values("GlobalKey")
+              .reset_index(drop=True)
+              .drop(columns=["id", "index"]))
```
thomasyu888, line 432: You're doing some operations here, is there a reason we wouldn't want to create a spark dataframe to compare in the assertion?

(I'm ok with this, was just curious)

tests/test_json_to_parquet.py
```diff
@@ -0,0 +1,531 @@
...
+                glue_context=glue_context,
+                workflow_run_id="testing"
+        )
+        s3_client = boto3.client("s3")
+        parquet_dataset = s3_client.list_objects_v2(
+                Bucket=artifact_bucket,
+                Prefix=parquet_key)
+        assert parquet_dataset["KeyCount"] > 0
+
+    def test_archive_existing_datasets(self, glue_context, artifact_bucket,
```
thomasyu888, line 471: For test readability, it'd be ideal if these are split into separate test cases:

* test_archive_existing_datasets_non_existent
* test_archive_existing_datasets_existent
* test_archive_existing_datasets_delete_upon_completion
* test_archive_existing_datasets_diff_workflow_id (appends)

This is true for many of these. It's not a hard and fast rule, but each test should ideally test a particular condition of a feature.  By doing that, it's also easier to know which part of the code base failed when a particular test fails.

tests/test_json_to_parquet.py
```diff
@@ -0,0 +1,531 @@
...
+                f"s3://{s3_bucket}"
+        )
+        s3_client.upload_file(
+                Filename=path,
+                Bucket=s3_bucket,
+                Key=s3_key
+        )
+
+    @pytest.fixture()
+    def glue_test_data(self, glue_flat_table_location, glue_nested_table_location,
```
thomasyu888, line 199: What are you trying to achieve here?  This seems like a job for `setup_method` for the test to execute this chunk of code for all the tests in the beginning.

thomasyu888 (commented):
Nice!  Will review in more detail soon.

tests/test_json_to_parquet.py
philerooski: This is pytest's way of doing setup/teardown for fixtures. In this fixture I define the workspace (S3) location and scope it to the class. When the class (e.g., the entire test suite for JSON to Parquet) completes, I clean up the workspace using the post-yield code.

tests/test_json_to_parquet.py
philerooski: They can be, but why would I want to use yield?

tests/test_json_to_parquet.py
philerooski: If there is an issue on AWS end with the crawler I don't want the tests to hang forever. This is a timeout. Using a while loop is also possible but would require additional logic to both check for the condition and keep a counter for the timeout (`while crawler_not_done and my_counter < timout`)

tests/test_json_to_parquet.py
philerooski: I'm more familiar with doing this in pandas than in pyspark. It's not really clear to me if pyspark has the dataframe manipulation methods needed to directly compare two dataframes since it's not designed for that kind of thing.

rxu17: It's relatively a new thing but while researching for ETL-430 (which you'll see soon), I found out that `pyspark` has a `pandas on spark` object where they mimic the pandas API but for spark dataframes to make it easier for pandas users to scale up to spark without having to learn all of the spark syntax from the ground up. It used to be a standalone project called [koalas](https://koalas.readthedocs.io/en/latest/?badge=latest) but spark saw the benefits and decided to officially include it in `pyspark`.

See here for more info/a sneak peek:

- [Pandas on spark quickstart](https://spark.apache.org/docs/latest/api/python/reference/pyspark.pandas/index.html)
- [Pandas on spark API](https://spark.apache.org/docs/latest/api/python/reference/pyspark.pandas/index.html)

tests/test_json_to_parquet.py
```diff
@@ -0,0 +1,531 @@
...
+                    raise ValueError("Found test data that is neither flat nor nested.")
+                absolute_path = os.path.join(root, basename)
+                self.upload_test_data_to_s3(
+                        path=absolute_path,
+                        s3_bucket=artifact_bucket,
+                        table_location=this_table_location
+                )
+
+    @pytest.fixture(scope="class")
+    def glue_crawler_role(self, namespace):
```
rxu17, line 217: Could use moto here: https://docs.getmoto.org/en/latest/docs/services/iam.html

tests/test_json_to_parquet.py
```diff
@@ -0,0 +1,531 @@
...
+                PolicyArn=glue_service_policy_arn
+        )
+        iam_client.detach_role_policy(
+                RoleName=role_name,
+                PolicyArn=s3_read_policy_arn
+        )
+        iam_client.delete_role(RoleName=role_name)
+
+    @pytest.fixture()
+    def glue_crawler(self, glue_database, glue_database_name, glue_database_path, glue_flat_table,
```
rxu17, line 257: Could use moto here: https://docs.getmoto.org/en/latest/docs/services/glue.html

tests/test_json_to_parquet.py
rxu17, line 217: Could use moto here: https://docs.getmoto.org/en/latest/docs/services/iam.html

tests/test_json_to_parquet.py
```diff
@@ -0,0 +1,531 @@
...
+        """
+        pass
+
+    def test_run_glue_crawler(self, glue_crawler):
+        """
+        Create and run Glue crawler.
+        """
+        pass
+
+    def test_get_table(self, glue_database_name, glue_flat_table_name,
```
rxu17, line 355: Could use something like `@pytest.mark.parametrize` to separate out these
two tests and use the `ids` parameter to title what each test is testing so it will show up when you run pytest: https://docs.pytest.org/en/6.2.x/parametrize.html

tests/test_json_to_parquet.py
```diff
@@ -0,0 +1,531 @@
...
+        assert len(flat_table.schema().fields) == 2
+        nested_table = json_to_parquet.get_table(
+                table_name=glue_nested_table_name,
+                database_name=glue_database_name,
+                glue_context=glue_context
+        )
+        assert nested_table.count() == 1
+        assert len(nested_table.schema().fields) == 4
+
+    def test_has_nested_fields(self, glue_database_name, glue_flat_table_name,
```
rxu17, line 372: Could use something like `@pytest.mark.parametrize` to separate out these
two tests and use the ids parameter to title what each test is testing so it will show up when you run pytest: https://docs.pytest.org/en/6.2.x/parametrize.html

tests/test_json_to_parquet.py
```diff
@@ -0,0 +1,531 @@
...
+            assert not json_to_parquet.has_nested_fields(flat_table_schema)
+            nested_table = json_to_parquet.get_table(
+                    table_name=glue_nested_table_name,
+                    database_name=glue_database_name,
+                    glue_context=glue_context
+            )
+            nested_table_schema = nested_table.schema()
+            assert json_to_parquet.has_nested_fields(nested_table_schema)
+
+    def test_add_index_to_table(self, glue_database_name, glue_database_path,
```
rxu17, line 389: I see multiple asserts here for different components. Could separate into multiple tests and add into the name of the test function what it's testing for within `add_index_to_table`. Maybe some of these can be fixtures as I see it's dependent on each other.

tests/test_json_to_parquet.py
```diff
@@ -0,0 +1,531 @@
...
+                    "Parameters": {
+                        "classification": "json",
+                        "compressionType": "none",
+                        "typeOfData": "file",
+                    }
+                }
+        )
+        return glue_table
+
+    def upload_test_data_to_s3(self, path, s3_bucket, table_location):
```
rxu17, line 184: I think you could upload the test data with moto: https://docs.getmoto.org/en/latest/docs/services/s3.html and the rest of the functions can use that

rxu17 (commented):
Just doing a first pass here

tests/test_json_to_parquet.py
rxu17, line 1: I don't see tests here for `drop_deleted_healthkit_data`? Were those planned for an integration test instead?

tests/conftest.py
```diff
@@ -271,3 +271,26 @@ def pytest_addoption(parser):
...
+        help=("Indicates that tests are being run in the production account "
+              "and that the production cloudformation bucket ought to be used.")
+    )
+
+@pytest.fixture(scope="session")
+def namespace(pytestconfig):
+     yield pytestconfig.getoption("namespace")
+
+@pytest.fixture(scope="session")
+def artifact_bucket(pytestconfig):
```
rxu17, line 291: I'm wondering if it's safe to write test related things to our artifact bucket just because it's technically where our scripts for our etl workflow are found? Esp having the option of using the production artifact bucket. 

If this is mainly for integration testing (doing these actions in a physical bucket instead of mocking it), then would it be better for us to have a separate bucket for this use case?

tests/test_json_to_parquet.py
philerooski: I'm using fixtures rather than xunit-style `setup_method`. I had to navigate a lot of complexity here because we want to run these fixtures once per class (like `setup_method`) but we also wanted to use the other fixtures which these "setup" fixtures are dependent on in different tests. I can't pass a fixture to `setup_method`, so using `setup_method` would require maintaining a lot of redundant set-up logic.

Unfortunately, one of the third-party fixtures we use (`datadir`) is scoped at the function level, and function-level fixtures cannot be used in what ought to be class-level fixtures (e.g., `glue_test_data`, `glue_crawler`), so to get around this these fixtures are downgraded to function-level fixtures and each are passed to their own test at the beginning of the class (`test_upload_test_data` and `test_run_glue_crawler`), which is functionally equivalent to `setup_method` but gives us the advantage of being able to use reusable fixtures in these "setup" fixtures.

tests/test_json_to_parquet.py
philerooski: This is for setting up the test environment on AWS (e.g., uploading test data, creating Glue resources, running crawler to add partitions to Glue tables) so we don't want to use a mock here.

tests/test_json_to_parquet.py
philerooski: Likewise previous comment.

tests/test_json_to_parquet.py
philerooski: I think it's possible but since we are testing two different things (that both `table.count` is correct and `len(table.schema().fields` is correct) it would be a bit hacky and I'd rather use this straightforward, redundant way of testing then use an abstracted but hacky way.

tests/test_json_to_parquet.py
philerooski: Yes, that works. I changed it to use `@pytest.mark.parametrize`.

tests/test_json_to_parquet.py
thomasyu888: `yield` can be used as a way to "setup" and "teardown" resources.  Just a minor thing.

tests/test_json_to_parquet.py
thomasyu888: TIL: This, very neat.

tests/test_json_to_parquet.py
thomasyu888: Ah, I see. Thanks for the explanation!

tests/test_json_to_parquet.py
philerooski: The `tables_with_index` dict which caches the results of each call to `json_to_parquet.add_index_to_table` is passed to subsequent calls to `json_to_parquet.add_index_to_table`. So these cases are more than dependent on the same input (like `nested_table_relationalized`), they are nested. I.e., they're dependent on the results of one another. It's still possible to split them up, but it takes the same code and splits it into different fixtures which makes it less readable imo.

tests/test_json_to_parquet.py
thomasyu888: Got it. This makes sense!

tests/test_json_to_parquet.py
philerooski: I forgot to write that test.

tests/test_json_to_parquet.py
philerooski: I'm uncertain if moto can do what I need it to do in these tests. I borrowed this method of testing wherein I don't mock some AWS resources from Bridge Downstream. Back then, Moto had even less support for Glue than it does now and I wanted to do some complicated things that I still don't fully understand -- particularly, updating the partitions and metadata in the Glue table so that when the JSON to Parquet job calls `glue_context.create_dynamic_frame.from_catalog` we actually get data. I know how to do that by running the crawler, but I spent a few days trying to do it manually a while back and couldn't figure out how to do it without running the crawler first. So it's still a bit of a mystery to me. And if I can't read the mocked tables then moto is useless.

I think a moto rewrite could be in the cards for the future but since we already have a working solution I'm reluctant to spend a day exploring what could be an unfruitful way to potentially shave a few minutes off test runtime.

tests/conftest.py
philerooski: I'm not worried about it since every S3 reference uses the `workspace_key_prefix`, which is formatted like `{namespace}/tests/test_json_to_parquet`.

I'm actually going to remove the `--prod` optional flag because I don't think it's a plausible use case to run unit tests in the production environment.

tests/test_json_to_parquet.py
rxu17: Yeah I can see that it would look very hacky... Maybe we could separate it out into two different tests then and update the test function names? I'm just thinking of when this script grows and we add more tests or in the future if we haven't touched this in a while, and we get a test report and there are failing tests, it would be good to have the nested table and regular table be two separate tests here so they will be concurrently run instead of now where if the regular table test fails, we won't know if nested table test would fail or pass.

tests/test_json_to_parquet.py
philerooski: I split these up.

tests/test_json_to_parquet.py
philerooski: I split these up.

thomasyu888 (approved):
💯 Nice - its great to see these tests added.  Going to pre-approve

tests/test_json_to_parquet.py
rxu17: I think it might be that you have to merge `glue_database` and `test_glue_database` so they are in the same function.

philerooski: Yes, this seems to work. Thank you!

philerooski: Unfortunately that wasn't the hard part. Even if I mock all the AWS resources, all the actual work is being done with Dynamic Frames, meaning I still need to mock `GlueContext` (and `SparkContext`?) and `DynamicFrame`. The Context objects do their work within a Spark environment and I don't know what I can realistically mock. If I'm going to finish these tests by the end of this sprint I need to go with the current "dirty" approach rather than attempt a pure unit test style approach.

tests/test_json_to_parquet.py
philerooski: I added this test.

PR #59, opened 2023-06-24
[ETL-327] Integrate RECOVER tests into CI/CD

rxu17:
**Purpose:** This integrates all of our RECOVER tests into our github workflow `upload-and-deploy`. This composes of a some jobs:

`nonglue-unit-tests`:
- One job runs the non-glue tests under Python 3.9

`pytest-docker`:
- One job creates and uploads the dockerfile for AWS glue 3.0
- One job creates and uploads the dockerfile for AWS glue 4.0

`glue-unit-tests`:
- One job runs the tests under AWS glue 3.0
- One job runs the tests under AWS glue 4.0

Some outstanding questions are: 

- [x] - In order to use the `matrix` method for creating and launching the jobs for the two Dockerfiles (because I have to pass the account ID as part of the job output), I had to unmask the AWS account ID (it would be hidden prior), is that alright?
- [x] - It doesn't appear that we have synapse folders for pre-ETL or post-ETL for `dev`, do we plan to?
Created a temporary test dev folder for these: `syn51079325`
- [x] - We should have an integration test (like in BD) that triggers the S3 to JSON lambda with the pilot datasets to run the rest of the ETL for a `feature branch` so that `compare_parquet_datasets` can eventually run on our development pipeline. But if that is successful, and we merge the `feature branch` into `develop/main`, we should have a process after stack deployment to clear the main S3 bucket namespace on `develop` and trigger the S3 to JSON lambda to run the ETL on `develop/main`?
Yes, this process works. Json to parquet already has a process in place to automatically archive old datasets and clear the workspace.

Depends on #60

### DIFF
--- a/.github/workflows/upload-and-deploy.yaml
+++ b/.github/workflows/upload-and-deploy.yaml
@@ -8,6 +8,7 @@ on:
 env:
   NAMESPACE: main
   PYTHON_VERSION: 3.9
+  DEV_INPUT_DATA_BUCKET: recover-dev-input-data
 
 jobs:
 
@@ -46,7 +47,17 @@ jobs:
         run: echo "NAMESPACE=$GITHUB_REF_NAME" >> $GITHUB_ENV
 
       - name: Copy files to templates bucket, use dev cloudformation bucket
-        run: python src/scripts/manage_artifacts/artifacts.py --upload --namespace $NAMESPACE --cfn_bucket ${{ vars.CFN_BUCKET }}
+        run: >
+          python src/scripts/manage_artifacts/artifacts.py
+          --upload
+          --namespace $NAMESPACE
+          --cfn_bucket ${{ vars.CFN_BUCKET }}
+
+      - name: Copies over test files from ingestion bucket
+        run: >
+          aws sync s3://recover-dev-ingestion/ s3://$DEV_INPUT_DATA_BUCKET/$NAMESPACE/
+          --exclude "owner.txt"
+          --exclude "test_upload.rtf"
 
 
   nonglue-unit-tests:
@@ -70,9 +81,13 @@ jobs:
         run: |
           pipenv run python -m pytest tests/test_s3_event_config_lambda.py -v
           pipenv run python -m pytest tests/test_s3_to_glue_lambda.py -v
+
       - name: Test synapse folders for STS access with pytest
-        run: |
-          pipenv run python -m pytest tests/test_setup_external_storage.py --test-synapse-folder-id syn51084525 --test-ssm-parameter synapse-recover-auth -v
+        run: >
+          pipenv run python -m pytest tests/test_setup_external_storage.py
+          --test-synapse-folder-id syn51084525
+          --test-ssm-parameter synapse-recover-auth
+          -v
 
 
   pytest-docker:
@@ -132,6 +147,7 @@ jobs:
       ecr-username: ${{ steps.login-ecr.outputs[steps.ecr.outputs.username-key] }}
       ecr-password: ${{ steps.login-ecr.outputs[steps.ecr.outputs.password-key] }}
 
+
   glue-unit-tests:
     name: Run Pytest unit tests for AWS glue
     needs: pytest-docker
@@ -178,13 +194,15 @@ jobs:
 
       - name: Run Pytest unit tests under AWS 4.0
         if: matrix.tag_name == 'aws_glue_4'
-        run: su - glue_user --command "cd $GITHUB_WORKSPACE && python3 -m pytest tests/test_json_to_parquet.py --namespace $NAMESPACE -v"
+        run: >
+          su - glue_user --command "cd $GITHUB_WORKSPACE &&
+          python3 -m pytest tests/test_json_to_parquet.py --namespace $NAMESPACE -v"
 
 
   sceptre-deploy-develop:
     name: Deploys branch using sceptre
     runs-on: ubuntu-latest
-    needs: [pre-commit, upload-files, nonglue-unit-tests, pytest-docker, glue-unit-tests]
+    needs: [upload-files, nonglue-unit-tests, glue-unit-tests]
     environment: develop
     # These permissions are needed to interact with GitHub's OIDC Token endpoint.
     permissions:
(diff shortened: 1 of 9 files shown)


.github/workflows/upload-and-deploy.yaml
```diff
@@ -46,13 +47,162 @@ jobs:
...
-        run: python src/scripts/manage_artifacts/artifacts.py --upload --namespace $NAMESPACE --cfn_bucket ${{ vars.CFN_BUCKET }}
+        run: >
+          python src/scripts/manage_artifacts/artifacts.py
+          --upload
+          --namespace $NAMESPACE
+          --cfn_bucket ${{ vars.CFN_BUCKET }}
+
+      - name: Copies over test files from ingestion bucket
+        run: >
+          aws s3 sync s3://recover-dev-ingestion/ s3://$DEV_INPUT_DATA_BUCKET/$NAMESPACE/
```
philerooski, line 58: We should move the pilot data to a specific prefix in the ingestion bucket (`s3://recover-dev-ingestion/pilot-data`? I'll leave the naming convention up to you). Digital health can also write to this bucket so we want to keep the behavior of this line predictable.

.github/workflows/upload-and-deploy.yaml
```diff
@@ -46,13 +47,162 @@ jobs:
...
+            dockerfile: tests/Dockerfile.aws_glue_4
+    environment: develop
+    steps:
+      - name: Assume AWS role
+        uses: aws-actions/configure-aws-credentials@v2
+        with:
+          role-to-assume: ${{ vars.AWS_CREDENTIALS_IAM_ROLE }}
+          aws-region: "us-east-1"
+          # unmasking of the AWS account ID allows the acct id to pass through outputs
+          mask-aws-account-id: "no"
```
philerooski, line 118: Why do we need to pass the account ID through outputs?

.github/workflows/upload-and-deploy.yaml
```diff
@@ -88,10 +238,53 @@ jobs:
           Payload: '{"RequestType": "Create"}'
           LogType: Tail
 
+
+  integration-test-develop:
```
philerooski, line 242: Did you check that this runs through the pipeline as expected? No failures and you can read the resulting parquet?

tests/README.md
```diff
@@ -41,17 +50,31 @@ cd <repo name>
...
+##### Running tests for json to parquet
+
+You will need to specify a namespace variable when running the json to parquet tests so that
+the AWS resources are created and run under this namespace.
+You can find your test related resources here (for reviewing/troubleshooting):
+
+`s3://<recover_artifacts_bucket_name>/<namespace>/tests/test_json_to_parquet`
+
+```shell script
+python3 -m pytest <path_to_your_specific_script> -namespace <name_of_namespace_to_use> -v
```
philerooski, line 62: `--namespace`?

philerooski (approved):

.github/workflows/upload-and-deploy.yaml
rxu17: So BD's account ID is also unmasked.
Essentially because we need to pass `ecr-registry` from the job `pytest-docker` whose value is `steps.login-ecr.outputs.registry` which contains the account ID, through so `glue-unit-tests` job can use it to get the docker container path

.github/workflows/upload-and-deploy.yaml
rxu17: Yes, I finally got the integration test part to actually trigger the lambda. The only thing is I'm `sam local` building and invoking our lambda. I'm not sure if we want to trigger our lambda directly using `aws invoke` or some github action that we already use? In order to do so, have to update our s3 to glue lambda template and stack configs because we currently don't have a set s3 to glue lambda function name

tests/README.md
```diff
@@ -4,26 +4,35 @@ Tests are defined in the `tests` folder in this project.
...
 
-Run the following commands to run tests for the s3_to_json script (in develop).
+These scripts needs to be run inside a Docker container:
+
+- Under AWS 3.0 (Dockerfile.aws_glue_3)
+  - test_compare_parquet_datasets.py
+  - test_s3_to_json.py
+
+- Under AWS 4.0 (Dockerfile.aws_glue_4)
+  - test_json_to_parquet.py (Note that these tests deploys test resources to aws and will take several min to run)
```
thomasyu888, line 19: Will there be a time when we can have one docker container?

thomasyu888 (approved):
🔥 Nice work here!  Going to pre-approve, just some minor comments.

.github/workflows/upload-and-deploy.yaml
rxu17: Yeah the output will be a bit different from BD's just because we use the `matrix` strategy so I had to tweak it a bit.

.github/workflows/upload-and-deploy.yaml
```diff
@@ -46,13 +48,171 @@ jobs:
...
+            dockerfile: tests/Dockerfile.aws_glue_3
+          - tag_name: aws_glue_4
+            dockerfile: tests/Dockerfile.aws_glue_4
+    environment: develop
+    steps:
+      - name: Assume AWS role
+        uses: aws-actions/configure-aws-credentials@v2
+        with:
+          role-to-assume: ${{ vars.AWS_CREDENTIALS_IAM_ROLE }}
+          aws-region: "us-east-1"
```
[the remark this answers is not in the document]
rxu17: It's not required I believe

tests/README.md
rxu17: Yes, when AWS python shell jobs support Python 3.10

.github/workflows/upload-and-deploy.yaml
```diff
@@ -46,13 +48,171 @@ jobs:
...
+        username: ${{ needs.pytest-docker.outputs.ecr-username }}
+        password: ${{ needs.pytest-docker.outputs.ecr-password }}
+      env:
+        DISABLE_SSL: true
+      options: "--user root"
+    steps:
+      - name: Assume AWS role
+        uses: aws-actions/configure-aws-credentials@v2
+        with:
+          role-to-assume: ${{ vars.AWS_CREDENTIALS_IAM_ROLE }}
```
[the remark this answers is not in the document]
rxu17: It's not required I believe

.github/workflows/upload-and-deploy.yaml
```diff
@@ -46,13 +48,171 @@ jobs:
...
+          aws-region: "us-east-1"
+      - uses: actions/checkout@v3
+      - run: chown -R glue_user $GITHUB_WORKSPACE
+      - run: su - glue_user --command "aws configure set aws_access_key_id $AWS_ACCESS_KEY_ID"
+      - run: su - glue_user --command "aws configure set aws_secret_access_key $AWS_SECRET_ACCESS_KEY"
+      - run: su - glue_user --command "aws configure set aws_session_token $AWS_SESSION_TOKEN"
```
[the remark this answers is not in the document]
rxu17: yeah, I'm not 100% sure if these lines are required as I copied them over from BD. When I looked up setting up ECR and using the containers from ECR via github actions, it seemed very particular about things, so I just wanted to keep it consistent with BD here since it's working there.

src/lambda_function/s3_to_glue/events/generate_test_event.py
```diff
@@ -16,8 +16,8 @@
 import json
 import boto3
 
-SINGLE_RECORD_OUTFILE = 'single-record.json'
-MULTI_RECORD_OUTFILE = 'records.json'
+SINGLE_RECORD_OUTFILE = './src/lambda_function/s3_to_glue/events/single-record.json'
+MULTI_RECORD_OUTFILE = './src/lambda_function/s3_to_glue/events/records.json'
```
philerooski, line 20: I don't like this change, actually. We shouldn't assume that this script will always be run from the repository root. If we run it anywhere else (like in the script directory) we'll end up with this unnecessarily deep folder nesting (`/src/lambda_function/s3_to_glue/events/src/lambda_function/s3_to_glue/events/single-record.json` for example). Why don't we make the output directory a command line argument instead? Or leave it as it is and run this script from the script directory in `upload-and-deploy` so that we can avoid the awkward `sam local invoke -e ../../..` syntax?

philerooski (changes requested):

src/lambda_function/s3_to_glue/events/generate_test_event.py
rxu17: Yeah the problem I ran into was that we had be inside the directory of the lambda app prior to running that command of `sam local invoke`, so it forced the `records.json` to be generated within that directory under `events`. 

I am in favor of the latter. This way for github actions, we'll be running the script from the repository root and if we need to run it outside of GH actions, we can specify where to save it.

philerooski (approved):
Can we squash the commits before merging?

PR #60, opened 2023-06-26
[ETL-483] Fix testing of bucket access via STS token credentials

rxu17:
**Purpose:** This PR fixes the test(s) in `test_setup_external_storage.py` so that it properly tests if STS access credentials can give you list bucket access to a STS-enabled synapse folder.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -77,7 +77,6 @@ access has been set for a given synapse folder (in develop).
 python3 -m pytest tests/test_setup_external_storage.py
 --test-bucket <put_bucket_name_here>
 --test-synapse-folder-id <put_synapse_folder_id_here>
---test-ssm-parameter <put_ssm_parameter_here_or_leave_blank>
 --namespace <put_namespace_here>
 --test_sts_permission <put_the_type_of_permission_to_test_here>
 ```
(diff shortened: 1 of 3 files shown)


tests/README.md
```diff
@@ -72,11 +72,12 @@ python3 -m pytest tests/test_s3_event_config_lambda.py -v
...
-
-- test-synapse-folder-id - synapse id of the folder to check STS access for
-- test-ssm-parameter - ssm parameter to get AWS credentials for otherwise leave blank and it will pull credentials from the environment
 
 ```shell script
-python3 -m pytest tests/test_setup_external_storage.py --test-synapse-folder-id <put_synapse_folder_id_here> --test-ssm-parameter <put_ssm_parameter_here_or_leave_blank>
+python3 -m pytest tests/test_setup_external_storage.py
+--test-bucket <put_bucket_name_here>
+--test-synapse-folder-id <put_synapse_folder_id_here>
+--test-ssm-parameter <put_ssm_parameter_here_or_leave_blank>
```
philerooski, line 80: I see that `--test-ssm-parameter` is not a new parameter for this PR, but what do you think about having this fixture return a hardcoded string literal in conftest.py, rather than passing this on the command line? We only have one SSM parameter and it doesn't change very often.(If it does, we can update the hard coded string literal.)

philerooski (approved):

tests/README.md
rxu17: Sure! That's fine with me

PR #61, opened 2023-06-30
[ETL-407] Add integration tests for prod/staging

philerooski:
Submits all production data to the staging namespace's pipeline if the github ref is main.

### DIFF
--- a/.github/workflows/upload-and-deploy.yaml
+++ b/.github/workflows/upload-and-deploy.yaml
@@ -10,6 +10,7 @@ env:
   PYTHON_VERSION: 3.9
   DEV_INPUT_BUCKET: recover-dev-input-data
   DEV_PROCESSED_BUCKET: recover-dev-processed-data
+  PROD_INPUT_BUCKET: recover-input-data
 
 jobs:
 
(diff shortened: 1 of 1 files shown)


thomasyu888 (approved):
🔥 awesome!

.github/workflows/upload-and-deploy.yaml
```diff
@@ -313,3 +315,41 @@ jobs:
 
       - name: Deploy sceptre stacks to staging on prod
         run: pipenv run sceptre --var "namespace=staging" launch prod --yes
+
+
+  integration-test-staging:
```
rxu17, line 320: Should we have a section for `if: github.ref_name == 'main' since we'd be only deploying when a feature gets merged into main?

rxu17, line 320: Should we have a section for `if: github.ref_name == 'main'` since we'd be only deploying when a feature gets merged into main?

rxu17 (approved):
LGTM! Just one commetn

.github/workflows/upload-and-deploy.yaml
rxu17: Nevermind, it requires `sceptre-deploy-staging` which already has that

.github/workflows/upload-and-deploy.yaml
```diff
@@ -10,6 +10,8 @@ env:
   PYTHON_VERSION: 3.9
   DEV_INPUT_BUCKET: recover-dev-input-data
   DEV_PROCESSED_BUCKET: recover-dev-processed-data
+  PROD_INPUT_BUCKET: recover-input-data
+  PROD_PROCESSED_BUCKET: recover-processed-data
```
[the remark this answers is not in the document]
philerooski: No, I was just mirroring the lines above. Probably we will never need it.

PR #62, opened 2023-07-05
[ETL-490] Delete local copy of JSON after uploading to S3

philerooski:


tests/test_s3_to_json.py
```diff
@@ -105,6 +105,7 @@ def test_write_healthkitv2samples_file_to_json_dataset(self, s3_obj, namespace,
                 dataset_identifier="HealthKitV2Samples",
                 metadata=sample_metadata["Metadata"],
                 workflow_run_properties=workflow_run_properties,
+                delete_upon_successful_upload=False,
```
rxu17, line 108: Do you think we need a test that tests that `os.remove` does get called as the default for `write_file_to_json_dataset` just to make sure that's always true?

rxu17 (approved):
LGTM! Just one comment

tests/test_s3_to_json.py
philerooski: Yeah, I'll add it.

src/glue/jobs/s3_to_json.py
```diff
@@ -21,7 +21,8 @@ def write_file_to_json_dataset(
         json_path: str,
         dataset_identifier: str,
         metadata: dict,
-        workflow_run_properties: dict) -> str:
+        workflow_run_properties: dict,
+        delete_upon_successful_upload=True) -> str:
```
thomasyu888, line 25: Nit: add type hint.

`bool = True`

thomasyu888 (approved):
💯 this is great to have! Thanks!

PR #63, opened 2023-07-06
[ETL-492] Update Glue table schemas to match crawler findings

philerooski:


rxu17 (approved):
LGTM! Will this schema change for the production data affect test runs with the pilot data given that the pilot data doesn't have these updated fields/new fields?

src/glue/resources/table_columns.yaml
rxu17, line 1: Will this schema change for the production data affect test runs with the pilot data given that the pilot data doesn't have these updated fields/new fields?

thomasyu888 (approved):
🔥 I'm going to pre-approve but I have the same question as rixing.  Thanks for pushing this through!

src/glue/resources/table_columns.yaml
philerooski: The changes here fall into a few different categories:

* New fields -- these changes don't affect the pilot data. The fields aren't in the data, so they aren't included in the parquet.
* Overdue changes -- In some cases, specifically in `EnrolledParticipants.EventDates` and `SymptomLog.Value`, we hadn't updated the types after the transformations we do to these fields in the S3 to JSON job. They _are_ string in the original data, but they're string-encoded JSON. We load them as dicts and write them as JSON objects in the ndjson, so the schemas reflect the object structure now.

So the pilot data isn't affected by these changes. It's fair to ask why was nothing breaking before because of the overdue changes? TBH, I don't think Glue entirely respects the data types we use in our tables. In some cases it just uses whatever data type it detects and can load the data with.

philerooski: @thomasyu888 see my comment above

PR #65, opened 2023-07-25
[ETL-505] Add Garmin transforms to S3 to JSON job

philerooski:
In addition to the Garmin transforms, I refactored the S3 to JSON job to be more functional. What used to just be
* `write_file_to_json_dataset`

has been split up into three different functions:
* `transform_json` - which takes care of any inserts/updates to the JSON data
* `transform_object_to_array_of_objects` - which is a helper function for `transform_json`
* `get_output_filename` - which takes care of filename / S3 object name formatting.

As a result of refactoring in the job, the tests have been refactored as well. Some tests were deleted and replaced with a more specific test.

### DIFF
--- a/src/glue/jobs/s3_to_json.py
+++ b/src/glue/jobs/s3_to_json.py
@@ -16,81 +16,232 @@ import zipfile
 import boto3
 from awsglue.utils import getResolvedOptions
 
-def write_file_to_json_dataset(
-        z: zipfile.ZipFile,
-        json_path: str,
-        dataset_identifier: str,
-        metadata: dict,
-        workflow_run_properties: dict,
-        delete_upon_successful_upload: bool=True) -> str:
+DATA_TYPES_WITH_SUBTYPE = ["HealthKitV2Samples", "HealthKitV2Statistics"]
+
+logging.basicConfig(level=logging.DEBUG)
+logger = logging.getLogger(__name__)
+
+def transform_object_to_array_of_objects(
+        json_obj_to_replace: dict,
+        key_name: str,
+        key_type: type,
+        value_name: str,
+        value_type: type,) -> list:
     """
-    Write a JSON from a zipfile to a JSON dataset.
+    Transforms a dictionary object into an array of dictionaries with specified
+    key and value types.
 
-    Metadata fields derived from the file basename are inserted as top-level fields
-    and the NDJSON is written to a JSON dataset in S3.
+    This function takes a dictionary object `json_obj_to_replace` and transforms
+    it into an array of dictionaries, where each dictionary in the array contains
+    two key-value pairs. The keys are specified by the `key_name` and `value_name`
+    parameters, and the values are converted to the corresponding types specified
+    by the `key_type` and `value_type` parameters.
 
-    Args:
-        z (zipfile.Zipfile): The zip archive as provided by the data provider.
-        json_path (str): A JSON path relative to the root of `z`.
-        dataset_identifier (str): The data type of `json_path`.
-        metadata (dict): Metadata derived from the file basename.
-        workflow_run_properties (dict): The workflow arguments
-        delete_upon_successful_upload (bool): Whether to delete the local
-            copy of the JSON file after uploading to S3. Set to False
-            during testing.
+    Parameters:
+        json_obj_to_replace (dict): The input dictionary object to be
+            transformed into an array of dictionaries.
+        key_name (str): The name of the key in the output dictionaries.
+        key_type (type): The type to which the values corresponding to `key_name`
+            should be converted in the output dictionaries.
+        value_name (str): The name of the value in the output dictionaries.
+        value_type (type): The type to which the values corresponding to `value_name`
+            should be converted in the output dictionaries.
 
     Returns:
-        output_path (str) The local path the file was written to.
+        list: An array of dictionaries, where each dictionary contains
+            two key-value pairs:
+              - The key specified by `key_name` with the value converted to the type specified by
+                `key_type`.
+              - The value specified by `value_name` with the value converted to the type specified
+                by `value_type`.
+
+    Examples:
+        json_obj = {'TimeOffsetHeartRateSamples': {"0": 62, "1": 63, "2": 62}}
+        transformed_array = transform_object_to_array_of_objects(
+            json_obj,
+            key_name='OffsetInSeconds',
+            key_type=int,
+            value_name='HeartRate',
+            value_type=int
+        )
+
+        # Resulting `transformed_array`:
+        # [
+        #     {'OffsetInSeconds': 0, 'HeartRate': 62},
+        #     {'OffsetInSeconds': 1, 'HeartRate': 63},
+        #     {'OffsetInSeconds': 2, 'HeartRate': 62},
+        # ]
     """
-    s3_client = boto3.client("s3")
-    logger = logging.getLogger(__name__)
-    os.makedirs(dataset_identifier, exist_ok=True)
-    s3_metadata = metadata.copy()
-    if s3_metadata["start_date"] is None:
-        s3_metadata.pop("start_date")
+    array_of_obj = []
+    value_error = "Failed to cast %s to %s"
+    for k, v in json_obj_to_replace.items():
+        try:
+            key_value = key_type(k)
+        except ValueError:
+            logger.warning(value_error, k,key_type)
+            key_value = None
+        try:
+            value_value = value_type(v)
+        except ValueError:
+            logger.warning(value_error, v, value_type)
+            value_value = None
+        obj = {
+            key_name: key_value,
+            value_name: value_value
+        }
+        array_of_obj.append(obj)
+    return array_of_obj
+
+def transform_json(
+        json_obj: dict,
+        dataset_identifier: str,
+        metadata: dict,) -> dict:
+    """
+    Perform the following transformations:
+
+    For every JSON:
+        - Add an export_start_date property (may be None)
+        - Add an export_end_date property (may be None)
+
+    For JSON whose data types have a subtype:
+        - Add subtype as "Type" property
+
+    For JSON whose data type is "EnrolledParticipants":
+        - Cast "CustomFields.Symptoms" and "CustomFields.Treatments" property
+          to an array. (May be an empty array).
+
+    For relevant Garmin data types:
+        - Some Garmin data types have one or more properties which are objects
+          (usually mapping time to some metric), that would be better formatted
+          as an array of objects. We transform these properties into arrays of
+          objects.
+
+    Args:
+        json_obj (str): A JSON object sourced from the JSON file of this data type.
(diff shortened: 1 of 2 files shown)


src/glue/jobs/s3_to_json.py
```diff
@@ -16,6 +16,237 @@
...
+        # Resulting `transformed_array`:
+        # [
+        #     {'OffsetInSeconds': 0, 'HeartRate': 62},
+        #     {'OffsetInSeconds': 1, 'HeartRate': 63},
+        #     {'OffsetInSeconds': 2, 'HeartRate': 62},
+        # ]
+    """
+    array_of_obj = [
+            {
+                key_name: key_type(k),
```
thomasyu888, line 74: Will there ever be cases where you won't be able to cast the value?

Things like (int('')) might cause an error.

src/glue/jobs/s3_to_json.py
```diff
@@ -16,6 +16,237 @@
...
+
+    Args:
+        json_obj (str): A JSON object sourced from the JSON file of this data type.
+        dataset_identifier (str): The data type of `json_obj`.
+        metadata (dict): Metadata derived from the file basename.
+
+    Returns:
+        json_obj (dict) The JSON object with the relevant transformations applied.
+    """
+    logger = logging.getLogger(__name__)
```
thomasyu888, line 113: Why not set the logger globally?

src/glue/jobs/s3_to_json.py
```diff
@@ -16,6 +16,237 @@
...
+    if metadata["start_date"] is not None:
+        json_obj["export_start_date"] = metadata["start_date"].isoformat()
+    else:
+        json_obj["export_start_date"] = None
+    json_obj["export_end_date"] = metadata.get("end_date").isoformat()
```
thomasyu888, line 118: Is there always an end date? Does .isoformat() work when the value is None?

src/glue/jobs/s3_to_json.py
```diff
@@ -16,6 +16,237 @@
...
+    if dataset_identifier in DATA_TYPES_WITH_SUBTYPE:
+        # This puts the `Type` property back where Apple intended it to be
+        json_obj["Type"] = metadata["subtype"]
+    if dataset_identifier == "SymptomLog":
```
thomasyu888, line 122: Can these be else ifs?

src/glue/jobs/s3_to_json.py
```diff
@@ -16,6 +16,237 @@
...
+        # Load JSON string as dict
+        json_obj["Value"] = json.loads(json_obj["Value"])
+    if dataset_identifier == "EnrolledParticipants":
+        for field_name in ["Symptoms", "Treatments"]:
+            if (
+                    field_name in json_obj["CustomFields"]
+                    and isinstance(json_obj["CustomFields"][field_name], str)
+               ):
+                if len(json_obj["CustomFields"][field_name]) > 0:
+                    # This JSON string was written in a couple different ways
```
thomasyu888, line 132: Wow, this is nasty. Is this worth letting care evolution know about this?

src/glue/jobs/s3_to_json.py
```diff
@@ -16,6 +16,237 @@
...
+                    except json.JSONDecodeError as e:
+                        # If it's not propertly formatted JSON, then we
+                        # can't read it, and instead store an empty list
+                        logger.warning(f"Problem CustomFields.{field_name}: "
+                                       f"{json_obj['CustomFields'][field_name]}")
+                        logger.warning(str(e))
+                        json_obj["CustomFields"][field_name] = []
+                else:
+                    json_obj["CustomFields"][field_name] = []
+    garmin_transform_types = {
```
thomasyu888, line 150: I like what you've done here, but is it higher risk in maintenance down the line if we forget that the first tuple has to be the key and the second tuple is value?

src/glue/jobs/s3_to_json.py
```diff
@@ -16,6 +16,237 @@
...
+                if prop_name in json_obj:
+                    array_of_obj = transform_object_to_array_of_objects(
+                            json_obj_to_replace=json_obj[prop_name],
+                            key_name=key_name,
+                            key_type=key_type,
+                            value_name=value_name,
+                            value_type=value_type
+                    )
+                    json_obj[prop_name] = array_of_obj
+            if len(property_hierarchy) == 2:
```
thomasyu888, line 197: Nit: the logic under this if statement sort of resembles the logic above : can they be combined to simplify the code?

src/glue/jobs/s3_to_json.py
```diff
@@ -16,6 +16,237 @@
...
+        - start_date
+        - end_date
+
+    Args:
+        metadata (dict): Metadata derived from the file basename.
+
+    Return:
+        str: A formatted file name.
+    """
+    if metadata["type"] in DATA_TYPES_WITH_SUBTYPE:
```
thomasyu888, line 230: I know this was code that already existed but does datatypes with subtypes always have startdate?

tests/test_s3_to_json.py
```diff
@@ -83,117 +83,309 @@ def json_file_basenames_dict(self):
...
+                "end_date": datetime.datetime(2023, 1, 14, 0, 0),
+        }
+        transformed_json = s3_to_json.transform_json(
+                json_obj={},
+                dataset_identifier=sample_metadata["type"],
+                metadata=sample_metadata
+        )
+
+        assert (
+            sample_metadata["start_date"].isoformat()
```
thomasyu888, line 128: You may want to explicitly write out the isoformat(). This is because if ever (isoformat) has a bug, this test would pass.  (Highly unlikely but just something to think about when writing tests that use the same functions as the code you're testing.

tests/test_s3_to_json.py
```diff
@@ -83,117 +83,309 @@ def json_file_basenames_dict(self):
...
+        transformed_json = s3_to_json.transform_json(
+                json_obj={
+                    "Summaries": [
+                        {
+                            "EpochSummaries": epoch_summaries,
+                            "Dummy": 1
+                        },
+                        {
+                            "EpochSummaries": epoch_summaries,
+                            "Dummy": 1
```
thomasyu888, line 320: Should this dummy value be different?

tests/test_s3_to_json.py
```diff
@@ -83,117 +83,309 @@ def json_file_basenames_dict(self):
...
-        with zipfile.ZipFile(io.BytesIO(s3_obj["Body"])) as z:
-            output_file = s3_to_json.write_file_to_json_dataset(
-                z=z,
-                json_path="SymptomLog_20230112-20230114.json",
-                dataset_identifier="SymptomLog",
-                metadata=sample_metadata["Metadata"],
-                workflow_run_properties=workflow_run_properties,
-                delete_upon_successful_upload=False,
-            )
+        transformed_epoch_summaries = [
```
thomasyu888, line 292: This is the expected transform right?

src/glue/jobs/s3_to_json.py
philerooski: I don't know. This data comes from Garmin (I believe CareEvolution is just an intermediary, they don't touch this part of the data), so it depends on what processes Garmin has in place so that they aren't writing incorrect types. I didn't experience any issues with the pilot data.

src/glue/jobs/s3_to_json.py
philerooski: The end date is derived from the filename. It's either included (in bold) like GarminHealthSnapshotSummary_20230626-**20230627**.json or EnrolledParticipants_**20230627**.json

Care Evolution always formats their exports like this. So, yes, there is always an end date.

src/glue/jobs/s3_to_json.py
philerooski: Either this way or with else if's would work. The conditions are predicated on the data type and each condition is mutually exclusive. Potentially in the future we will add non-mutually exclusive conditions (like `if dataset_identifier == data_type` and then later in the code... `if property == value`), in which case we wouldn't want to use else if, or only one of the conditional blocks will be executed.

src/glue/jobs/s3_to_json.py
philerooski: I haven't seen the `[{\\\"id\\\": ...` format in the production data (we would log a warning if we ran into a `JSONDecodeError`), so whatever issues they had with the pilot data seem to have been resolved.

src/glue/jobs/s3_to_json.py
philerooski: This was something I did intentionally. It seems pretty difficult to write this in a way that can handle an arbitrary amount of nesting of the JSON property. I would need to write a recursive function. The `len == 1` case would be the base case and you can begin to imagine how the recursive case might look with the extra `for...` block in the `len == 2` case. I think a recursive implementation would be more difficult to understand/debug, so unless we need to handle deeper nesting in the future I think it's better to have it implemented more explicitly -- although a bit redundantly.

src/glue/jobs/s3_to_json.py
philerooski: Yes, so far.

tests/test_s3_to_json.py
philerooski: That's an interesting edge case that `isoformat` might have a bug, although it's from the python standard library. I don't think it's an edge case we need to test for. That would be up to python's developers to test for bugs in `datetime`. 

Perhaps a more likely scenario is that the default ISO format outputted by `isoformat` changes. I would be okay with that, as long as it's ISO standard.

tests/test_s3_to_json.py
philerooski: That's not really necessary, but I've removed `Dummy` from the second object anyways since we don't use it as part of an assert statement.

tests/test_s3_to_json.py
philerooski: Yes.

src/glue/jobs/s3_to_json.py
philerooski: That's fair. I've made it global now.

src/glue/jobs/s3_to_json.py
thomasyu888: That makes sense.  This leaves room open for future conditions.

src/glue/jobs/s3_to_json.py
thomasyu888: That makes sense.  Can you add a comment here about that?

tests/test_s3_to_json.py
thomasyu888: Makes sense to me!

src/glue/jobs/s3_to_json.py
thomasyu888: We just have to make sure we are monitoring the logs for these warnings.

thomasyu888 (approved):
💯 awesome!

PR #66, opened 2023-07-25
[ETL-499] JSON to Parquet support for Garmin data types

philerooski:
Blocked by #65 

Changes fall into three categories:

- Changes to `config/*` files - this replaces references to the S3 bucket where we store our templates so that we don't run into the file size limitation on the `--templateBody` parameter when the `aws cloudformation` tool validates the templates.
- Changes to `src/glue/jobs/json_to_parquet.py` - This adds the index fields for Garmin data types to our `INDEX_FIELD_MAP` global, adds additional logic to reference the `InsertedDate` field when dropping duplicates, and does a small amount of refactoring.
- Changes to `src/glue/resources/table_columns.yaml` - This adds the schema definitions for Garmin data types while reflecting the transformations we do for these data types in #65 .


src/glue/jobs/json_to_parquet.py
rxu17, line 1: Does there need to be updates to the tests to have examples with `InsertedDate` for this new change?

rxu17 (approved):

src/glue/jobs/json_to_parquet.py
```diff
@@ -115,11 +133,17 @@ def get_table(
                 table_name=table_name
             )
     )
+    if table.count() == 0:
+        return table
+    field_names = [field.name for field in table.schema().fields]
     spark_df = table.toDF()
+    if "InsertedDate" in field_names:
```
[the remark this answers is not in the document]
philerooski: There was an email with Care Evolution. Sorry I forgot to cc you! Here is their words:

> For files that contain "InsertedDate", the "InsertedDate" should be used to identify the most recent values and resolve duplicates.  The "InsertedDate" is included in those files because it is possible for there to be duplicates within the same export file.  For the files without InsertedDate, you should continue to use the date included in the filename to resolve duplicates.

src/glue/jobs/json_to_parquet.py
philerooski: Yes 🙃

PR #67, opened 2023-07-31
Add an owner.txt to the ingestion bucket

philerooski:
Synapse already has read/write access to the ingestion bucket, so the only thing to do to enable the use of the bucket as an external storage location is to write an owner.txt.

In develop, owner.txt is written to `main`.
In prod, owner.txt is written to the bucket root. This is where production data (e.g., `adults\/v1` and `pediatric\/v1`) already live.

### DIFF
--- a/config/develop/namespaced/s3-ingestion-bucket-owner-txt.yaml
+++ b/config/develop/namespaced/s3-ingestion-bucket-owner-txt.yaml
@@ -1,0 +1,11 @@
+template:
+  type: file
+  path: s3-owner-txt.yaml
+stack_name: "{{ stack_group_config.namespace }}-recover-dev-ingestion-bucket-owner-txt"
+dependencies:
+  - develop/cfn-s3objects-macro.yaml
+  - develop/s3-ingestion.yaml
+parameters:
+  BucketName: !stack_output_external recover-dev-ingestion-bucket::Bucket
+  SynapseIds: "3461799,3455604" # RecoverETL and synapse-service-sysbio-recover-data-storage-01
+  OwnerTxtKeyPrefix: {{ stack_group_config.namespace }}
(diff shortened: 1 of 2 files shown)


config/develop/namespaced/s3-ingestion-bucket-owner-txt.yaml
```diff
@@ -0,0 +1,11 @@
...
+  type: file
+  path: s3-owner-txt.yaml
+stack_name: "{{ stack_group_config.namespace }}-recover-dev-ingestion-bucket-owner-txt"
+dependencies:
+  - develop/cfn-s3objects-macro.yaml
+  - develop/s3-ingestion.yaml
+parameters:
+  BucketName: !stack_output_external recover-dev-ingestion-bucket::Bucket
+  SynapseIds: "3461799,3455604" # RecoverETL and synapse-service-sysbio-recover-data-storage-01
+  OwnerTxtKeyPrefix: {{ stack_group_config.namespace }}
```
rxu17, line 11: Was there a particular reason we are removing this for the prod version and having the owner.txt being written to the root of the bucket? i think the prod config for the input bucket has this

rxu17 (approved):
LGTM! Just one question

config/develop/namespaced/s3-ingestion-bucket-owner-txt.yaml
philerooski: We don't really control the organization of data in the production ingestion bucket. So far data has been written to two different folders at the bucket root (`adults\/v1` and `pediatric\/v1`), so unless we move the production data elsewhere, digital health will need to have access to the bucket root and update their scripts to copy data to that "elsewhere" folder. Having everything at the bucket root doesn't really bother me, because there is only one set of _production_ data. We might send that data through different namespaces (e.g., `main` and `staging`), but there isn't a separate set of production data for `main` and another for `staging`.

rxu17: Ah makes sense

PR #71, opened 2023-08-18
[ETL-521] Run JSON to Parquet jobs in groups

philerooski:
Reorganizes the primary workflow by running JSON to Parquet jobs in groups determined by their data type. Jobs are run in this order:

1. Non-device specific jobs (EnrolledParticipants, SymptomLog)
2. HealthKit jobs
3. Fitbit jobs
4. Google jobs
5. Garmin jobs

A job group is triggered if every job in the previously run job group succeeds. Structuring the JSON to Parquet jobs like this does increase the total workflow run time (the pilot data processing time increased from 13 minutes to 27 minutes) - since not as many jobs are being run in parallel. But it does put a lot less stress on our rate limits and we're less likely to get our jobs killed because we have exceeded the rate limit.

I've also increased the JSON to Parquet job default timeout to 12 hours from 2 hours.

### DIFF
--- a/templates/glue-job-JSONToParquet.j2
+++ b/templates/glue-job-JSONToParquet.j2
@@ -40,7 +40,7 @@ Parameters:
   TimeoutInMinutes:
     Type: Number
     Description: The job timeout in minutes (integer).
-    Default: 120
+    Default: 720
 
   TempS3Bucket:
     Type: String
--- a/templates/glue-workflow.j2
+++ b/templates/glue-workflow.j2
@@ -53,7 +53,8 @@ Resources:
   {% set datasets = [] %}
   {% for v in sceptre_user_data.dataset_schemas.tables.keys() if not "Deleted" in v %}
     {% set dataset = {} %}
-    {% do dataset.update({'table_name': 'dataset_' + v.lower()})%}
+    {% do dataset.update({'data_type': v}) %}
+    {% do dataset.update({'table_name': 'dataset_' + v.lower()}) %}
     {% do dataset.update({'stackname_prefix': '{}'.format(v.replace('_',''))}) %}
     {% do datasets.append(dataset) %}
   {% endfor %}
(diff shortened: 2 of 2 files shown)


templates/glue-job-JSONToParquet.j2
```diff
@@ -40,7 +40,7 @@ Parameters:
   TimeoutInMinutes:
     Type: Number
     Description: The job timeout in minutes (integer).
-    Default: 120
+    Default: 720
```
rxu17, line 43: Do we have an idea of how long the total workflow time is now (with the 2nd round of processing data)?

templates/glue-workflow.j2
```diff
@@ -88,11 +89,11 @@ Resources:
     Properties:
       Name: !Sub "${Namespace}-S3ToJsonCompleteTrigger"
       Actions:
-        {% for dataset in datasets %}
+        {% for dataset in datasets if not "HealthKit" in dataset["data_type"] and not "Fitbit" in dataset["data_type"] and not "Google" in dataset["data_type"] and not "Garmin" in dataset["data_type"] %}
         - JobName: !Sub ${Namespace}-{{ dataset["stackname_prefix"]}}-Job
           Arguments: {"--glue-table": {{ "{}".format(dataset["table_name"]) }} }
         {% endfor %}
-      Description: This trigger runs after completion of the S3 to JSON job.
+      Description: This trigger kicks off every JSON to Parquet job which is not associated with a device and runs after completion of the S3 to JSON job.
```
rxu17, line 96: Should we add to the description something regarding the order in which the data type jobs will run and that they will be running one by one?

rxu17 (approved):
LGTM!

templates/glue-job-JSONToParquet.j2
thomasyu888: Is there a rationale behind the order or was the order random?

philerooski: > Do we have an idea of how long the total workflow time is now (with the 2nd round of processing data)?
 
Not sure, but some of the data types take longer than 2 hours to process. I needed to increase this time since I was getting timeout errors.

philerooski:
@thomasyu888 

> Is there a rationale behind the order or was the order random?

My understanding is that all the data types are of equal importance. So I didn't triage any specific device / data type and the order is effectively random.

templates/glue-workflow.j2
philerooski: I updated the description of the glue-workflow template. Is that what you had in mind?

thomasyu888 (approved):
🔥 LGTM!  Thanks for pushing this through!

templates/glue-workflow.j2
rxu17: Yup, thanks for doing that!

PR #72, opened 2023-08-29
[ETL-523] Increase number of workers for types FitbitIntradayCombined and HeathKitV2Samples

philerooski:


rxu17 (approved):
LGTM! Do you think in the future we should start having a config file that contains which data types use which amount of resources, maximum running time, order of data types to runs, etc (I don't know currently how we could pull that info using cloud formation though)

philerooski:
> LGTM! Do you think in the future we should start having a config file that contains which data types use which amount of resources, maximum running time, order of data types to runs, etc (I don't know currently how we could pull that info using cloud formation though)

@rxu17 Yes, that's a good solution. We can parse such a file using Jinja. Right now our configuration is simple enough that we can rely on the "dumb" method of having two cloudformation parameters.

thomasyu888 (approved):
💯 nice!

PR #76, opened 2023-09-06
[ETL-533] Add `cohort` to data as partition

philerooski:
- Added `cohort` as a partition in our Glue table schemas.
- This requires us to include `cohort` as a property in our JSON. I also did some light refactoring around `get_part_path` in the S3 to JSON job.
- To add these partitions to our table, we need a Glue crawler. I've added one to crawl every data type.
- Updated test data to reflect the way data is organized in production (e.g., in `adults_v1` and `pediatric_v1` directories). Updated test event generation script so that it does not include these "directories" as individual events.
- Updated tests

### DIFF
--- a/src/glue/jobs/json_to_parquet.py
+++ b/src/glue/jobs/json_to_parquet.py
@@ -125,6 +125,7 @@ def get_table(
             glue_context.create_dynamic_frame.from_catalog(
                  database=database_name,
                  table_name=table_name,
+                 additional_options={"groupFiles": "inPartition"},
                  transformation_ctx="create_dynamic_frame"
             )
             .resolveChoice(
@@ -322,7 +323,8 @@ def write_table_to_s3(
             frame = dynamic_frame,
             connection_type = "s3",
             connection_options = {
-                "path": s3_write_path
+                "path": s3_write_path,
+                "partitionKeys": ["cohort"]
             },
             format = "parquet",
             transformation_ctx="write_dynamic_frame")
@@ -390,8 +392,9 @@ def add_index_to_table(
         logger.info(f"Adding index to {original_field_name}")
         parent_index = (parent_table
                 .select(
-                    [selectable_original_field_name] + INDEX_FIELD_MAP[table_data_type])
-                .distinct())
+                    ([selectable_original_field_name, "cohort"]
+                     + INDEX_FIELD_MAP[table_data_type])
+                ).distinct())
         this_index = parent_index.withColumnRenamed(original_field_name, "id")
         df_with_index = this_table.join(
                 this_index,
--- a/src/glue/jobs/json_to_parquet.py
+++ b/src/glue/jobs/json_to_parquet.py
@@ -1,14 +1,13 @@
 """
 This script runs as a Glue job and converts a collection of JSON files
 (whose schema is defined by a Glue table), to a parquet dataset partitioned by
-assessmentid / year / month / day. Additionally, if the table has nested data,
+cohort. Additionally, if the table has nested data,
 it will be separated into its own dataset with a predictable name. For example,
-the info table (derived from info.json) has a field called "files" which is an
-array of objects. We will write out two parquet datasets in this case, an `info`
-dataset and an `info_files` dataset.
+the healthkitv2heartbeat data type has a field called "SubSamples" which is an
+array of objects. We will write out two parquet datasets in this case, a `healthkitv2heartbeat`
+dataset and an `healthkitv2heartbeat_subsamples` dataset.
 
-Before writing our tables to parquet datasets, we add the recordid,
-assessmentid, year, month, and day to each record in each table.
+
 """
 
 import os
@@ -108,6 +107,8 @@ def get_table(
     Duplicate samples are dropped by referencing the `InsertedDate`
     field, keeping the more recent sample. For any data types which
     don't have this field, we drop duplicates by referencing `export_end_date`.
+    Additionally, we drop any superfluous partition_* fields which are
+    added by Glue.
 
     Args:
         table_name (str): The name of the Glue table.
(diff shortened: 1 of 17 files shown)


src/glue/jobs/s3_to_json.py
```diff
@@ -398,7 +409,11 @@ def write_file_to_json_dataset(
             os.remove(output_path)
     return uploaded_files
 
-def get_part_path(metadata: dict, part_number: int, dataset_identifier: str, touch: bool):
```
rxu17, line 401: Did you want to mention `partition` somewhere here either in the function name or the description?

src/glue/jobs/s3_to_json.py
```diff
@@ -544,8 +563,21 @@ def main() -> None:
                 Key = message["source_key"]
         )
         s3_obj["Body"] = s3_obj["Body"].read()
+        cohort = None
+        if "adults_v1" in message["source_key"]:
```
rxu17, line 567: Will we ever have v2 or more cohorts? Do we feel like this is something we should parse out?

src/glue/jobs/s3_to_json.py
```diff
@@ -544,8 +563,21 @@ def main() -> None:
...
+            cohort = "adults_v1"
+        elif "pediatric_v1" in message["source_key"]:
+            cohort = "pediatric_v1"
+        else:
+            logger.warning(
```
rxu17, line 572: Does this warning appear for you in the logs? When i was troubleshooting something with s3 to json, I was reading that because this isn't a spark job, the logging info doesn't appear in cloudtrail in any of the logs or error outputs.

I had to do something like:
```
logger = logging.getLogger()
logger.setLevel(logging.DEBUG)
handler = logging.StreamHandler(sys.stdout)
handler.setLevel(logging.DEBUG)
formatter = logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s")
handler.setFormatter(formatter)
logger.addHandler(handler)
```
To get the logs to show at least at the INFO and DEBUG levels. But maybe WARN level logs are outputted.

Since this seems like a pretty important warning, should we have a test to make sure this is always caught or escalate it to an exception?

rxu17 (approved):
LGTM, some comments!

src/glue/jobs/json_to_parquet.py
```diff
@@ -322,7 +326,8 @@ def write_table_to_s3(
             frame = dynamic_frame,
             connection_type = "s3",
             connection_options = {
-                "path": s3_write_path
+                "path": s3_write_path,
+                "partitionKeys": ["cohort"]
```
thomasyu888, line 330: Do you envision us adding more partitions in the future?  If so, does it make sense to add them sooner than later?

src/glue/jobs/s3_to_json.py
thomasyu888: Nit: Just trying to think ahead in the future, but we might want to make all logs we put into cloudtrail machine parsable.  We can then use mine the logs for issues on a regular basis

thomasyu888 (approved):
🔥 This is great to add the cohorts as partition.  I'm going to pre-approve, but I agree with the review that @rxu17 provided. (great review!)

src/glue/jobs/json_to_parquet.py
```diff
@@ -135,9 +137,11 @@ def get_table(
     )
     if table.count() == 0:
         return table
-    field_names = [field.name for field in table.schema().fields]
     spark_df = table.toDF()
-    if "InsertedDate" in field_names:
+    for c in spark_df.columns:
+        if "partition_" in c: # superfluous field added by Glue
```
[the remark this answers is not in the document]
philerooski: No, this seems to be an artifact of loading the data as a DynamicFrame (perhaps from the `additional_options={"groupFiles": "inPartition"}` parameter?). There is no `partition_cohort` field in the Glue table schemas, but there is one after loading the data.

src/glue/jobs/s3_to_json.py
philerooski: No, these "parts" aren't actually related to the `cohort` partition. Although it is true that we are partitioning the data by writing it in parts, that terminology has the potential of becoming confounded with the hive-style partitions we get when writing the data to S3.

Here's a diagram of how this looks in S3:
- dataset_datatype
  * cohort=adults_v1 <- [ Hive partition ]
    - part0.snappy.parquet <- [ part "partition" ]
    - part1.snappy.parquet
    - ...
  * cohort=pediatric_v1
    - part0.snappy.parquet
    - part1.snappy.parquet
    - ...

src/glue/jobs/s3_to_json.py
rxu17: hmm is that something we would want some say over? The name of the S3 key prefix containing the cohort's data

src/glue/jobs/s3_to_json.py
philerooski: I was hoping to be productive today but I just spent the past 2 hours trying to figure out why this works 😓 

This logger flow helped my understanding: https://docs.python.org/3/howto/logging.html#logging-flow

So it seems like this would also work:
```
logger = logging.basicConfig(level=logging.DEBUG, stream=sys.stdout)
logger.handlers[0].setLevel(logging.DEBUG)
```
but it doesn't.

I also couldn't figure out why a `print` statement I had accidentally left in the `transform_object_to_array_of_objects` function would write to `/aws-glue/python-jobs/output` but a print statement I put at the global namespace wouldn't.

I ended up copying this code verbatim except for the format, which I replaced with the default format. Now I get my `debug` level statements (which I wasn't getting before), but I also get debug statements from every library used in the script, so now there is an overwhelming amount of logging.

Basically, there is no documentation about logging in an Glue python shell job and it is super confusing.

src/glue/jobs/json_to_parquet.py
philerooski: I'm not convinced that partitions are particularly useful right now. It's a way to keep cohort data separate from one another, but I'm not convinced that there's a lot of I/O performance benefits right now. Adult data is like 99% of our total data for one. Digital health also seems to be working with the entire dataset for their processing, so partitions probably aren't getting used.

src/glue/jobs/s3_to_json.py
philerooski: Not necessarily. It can be an arbitrary key prefix, and as long as it stays consistent we can filter for it and partition accordingly.

PR #77, opened 2023-09-07
[ETL-536] Modify lambda to poll SQS directly

rxu17:
**Purpose:** This modifies our S3 to JSON lambda function code to poll the SQS queue directly and grants any permissions in order to do so.
Our s3 to json lambda glue role has been updated to be namespaced to avoid affecting other development branches' glue roles as it changes the permissions to be SQS specific. 

**Future implications:**
This will make testing a bit different now. Previously we could just put a bunch of S3 event messages in JSON format and trigger our lambda with it directly through the console or `sam local invoke`. Now we have to put the specific S3 event messages in our SQS queue, and then trigger our lambda with any kind of test trigger event to poll from the SQS queue.

Note that the above affects [ETL-537](https://sagebionetworks.jira.com/browse/ETL-537) as our test events that trigger our Lambda are no longer S3 events but SQS events containing S3 events

[ETL-537]: https://sagebionetworks.jira.com/browse/ETL-537?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


philerooski:
>  Now we have to put the specific S3 event messages in our SQS queue, and then trigger our lambda with any kind of test trigger event to poll from the SQS queue

Why didn't you instead rewrite `generate_test_event.py` so that it sends the S3 event wrapped in an SQS event to the lambda directly? This is similar to what we already do in Bridge Downstream and would allow us to continue to send test events directly to the lambda.

rxu17:
> Why didn't you instead rewrite `generate_test_event.py` so that it sends the S3 event(s) wrapped in an SQS event to the lambda directly? This is similar to what we already do in Bridge Downstream and would allow us to continue to send test events directly to the lambda.

Because from the last recover discussion, it sounds like we would like to schedule the lambda at a certain time every day to poll the SQS queue for all messages and then proceed to process all of them. Because of this, lambda would be triggered by a cloudwatch trigger event instead of a SQS event with the S3 events thus we wouldn't be able to send test events directly to the lambda unless there is a way I'm not thinking of? 

So then modifying `generate_test_event.py` would be this ticket: [ETL-537](https://sagebionetworks.jira.com/browse/ETL-537)

philerooski:
> it sounds like we would like to schedule the lambda at a certain time every day to poll the SQS queue for all messages and then proceed to process all of them.

That might work, but it's not what I envisioned how the lambda would be triggered. I had thought we would use the SQS as an event source for the lambda. So the lambda would poll the SQS queue every 20 seconds (long polling) for new messages. Having received a new message, that's when our `BatchSize` and `MaximumBatchingWindowInSeconds` parameters come into play. The lambda will wait to trigger until it's either received the maximum number of messages it is willing to process in a batch or a certain amount of time has passed. 

>  lambda would be triggered by a cloudwatch trigger event instead of a SQS event with the S3 events thus we wouldn't be able to send test events directly to the lambda

Even if the lambda uses SQS as an event source, we can still send it events manually.

rxu17:
Based on what we discussed, we will adapt the event source mapping between the lambda and SQS queue, and get rid of the scheduler. We will set `MaximumBatchingWindowInSeconds` to the maximum time which is 5 minutes and have something in the lambda that checks if our PrimaryWorkflow is already running (and if it is, we send the messages to the DLQ, this is for another ticket)

Pipfile
```diff
@@ -17,3 +17,6 @@ pandas = "<1.5"
 moto = "~=4.1"
 datacompy = "~=0.8"
 docker = "~=6.1"
+pyhive = "~=0.7"
+thrift = "~=0.16"
+thrift-sasl = "~0.4.3"
```
thomasyu888, line 22: Can you speak to what these packages do? 

What Python versions are supported via these packages?

rxu17: My bad, I did not mean to push these. These are unused.

config/develop/namespaced/s3-to-glue-lambda.yaml
```diff
@@ -4,11 +4,14 @@ template:
...
 stack_name: '{{ stack_group_config.namespace }}-lambda-S3ToGlue'
 stack_tags: {{ stack_group_config.default_stack_tags }}
 parameters:
-  S3SourceBucketName: {{ stack_group_config.input_bucket_name }}
-  S3ToGlueRoleArn: !stack_output_external s3-to-glue-lambda-role::RoleArn
+  SQSQueueUrl: !stack_output_external "{{ stack_group_config.namespace }}-sqs-S3ToLambda::PrimaryQueueUrl"
+  SQSQueueArn: !stack_output_external "{{ stack_group_config.namespace }}-sqs-S3ToLambda::PrimaryQueueArn"
+  S3ToGlueRoleArn: !stack_output_external "{{ stack_group_config.namespace }}-s3-to-glue-lambda-role::RoleArn"
   PrimaryWorkflowName: !stack_output_external "{{ stack_group_config.namespace }}-glue-workflow::WorkflowName"
+  LambdaBatchSize: '10'
```
thomasyu888, line 16: Can we push these docs: https://docs.google.com/document/d/1K36ViBjzh32W32rA6mFzXEehyO2ee5qqAg2k7MHIF3g/edit#heading=h.urvxhb9crjdz to https://sagebionetworks.jira.com/wiki/spaces/IBC/pages/2733867066/Design+Documents

src/lambda_function/s3_to_glue/app.py
```diff
@@ -12,14 +12,56 @@
...
+    Args:
+        object_info (dict): Object information from source S3 bucket
+
+    Returns:
+        dict: returns back the object info dict if it passes the filter criteria
+            otherwise returns None
+    """
+    if not object_info["source_key"]:
+        logger.info(
+            "This object_info record doesn't contain a source key "
```
thomasyu888, line 33: nit: we may want to think about how these messages look like in cloudtrail.  We should be able to mine these messages easily to monitor this.

src/lambda_function/s3_to_glue/app.py
```diff
@@ -12,14 +12,56 @@
...
+    elif "owner.txt" in object_info["source_key"]:
+        logger.info(
+            f"This object_info record is an owner.txt and can't be processed.\nMessage: {object_info}"
+        )
+        return None
+    elif object_info["source_key"].endswith("/"):
+        logger.info(
+            f"This object_info record is a directory and can't be processed.\nMessage: {object_info}"
+        )
+        return None
```
thomasyu888, line 52: Nit: You may be able to simplify all these returns by just having an end return, BUT i like that you're being explicit here.

```
else:
    return object_info
return None
```

src/lambda_function/s3_to_glue/app.py
```diff
@@ -31,38 +73,56 @@ def submit_s3_to_json_workflow(messages: list[dict[str,str]], workflow_name):
...
+                objects_info=s3_objects_info,
+                workflow_name=os.environ["PRIMARY_WORKFLOW_NAME"],
+            )
+        # Delete the messages from the queue after all events are successfully submitted
+        for record in event["Records"]:
+            receipt_handle = record["receiptHandle"]
+            sqs.delete_message(
+                QueueUrl=os.environ["SQS_QUEUE_URL"], ReceiptHandle=receipt_handle
+            )
+        return {"statusCode": 200, "body": "All messages retrieved and processed."}
```
thomasyu888, line 125: Is there a "class" for the return or is this just the dict lambda expects?

src/lambda_function/s3_to_glue/app.py
```diff
@@ -31,38 +73,56 @@ def submit_s3_to_json_workflow(messages: list[dict[str,str]], workflow_name):
...
     )
 
 
-def lambda_handler(event, context):
-    """
-    The Lambda entrypoint
-
-    Given a list of files in s3 bucket to process, submit the keys and bucket
-    of the new files to a S3 to JSON Glue workflow.
+def lambda_handler(event, context) -> None:
```
thomasyu888, line 80: Nit: `-> dict`

src/lambda_function/s3_to_glue/app.py
```diff
@@ -31,38 +73,56 @@ def submit_s3_to_json_workflow(messages: list[dict[str,str]], workflow_name):
...
 
     Args:
-        event (dict): A fileview object
-        context (obj) : An object that provides methods and properties that
-        provides info about the invocation, function, and execution environment
-
-    Returns:
-        (None) Submits new file records to a Glue workflow.
+        event (json): SQS event
+        context: NA
```
thomasyu888, line 87: What is context?

src/lambda_function/s3_to_glue/app.py
```diff
@@ -31,38 +73,56 @@ def submit_s3_to_json_workflow(messages: list[dict[str,str]], workflow_name):
...
-        message = {"source_bucket": bucket_name, "source_key": object_key}
-        if "owner.txt" not in object_key:
-            messages.append(message)
-    if len(messages) > 0:
-        logger.info(
+    # Initialize SQS client
+    sqs = boto3.client("sqs")
+    s3_objects_info = []
+    if "Records" in event:
+        for record in event["Records"]:
```
thomasyu888, line 93: What is the scenario where "Records" isn't in the event dict?

src/lambda_function/s3_to_glue/app.py
```diff
@@ -31,38 +73,56 @@ def submit_s3_to_json_workflow(messages: list[dict[str,str]], workflow_name):
...
+                    object_key = s3_event["s3"]["object"]["key"]
+                    object_info = {
+                        "source_bucket": bucket_name,
+                        "source_key": object_key,
+                    }
+                    if filter_object_info(object_info) is not None:
+                        s3_objects_info.append(object_info)
+                    else:
+                        logger.info(
+                            f"Object doesn't meet the S3 event rules to be processed. Skipping\n{object_info}"
```
thomasyu888, line 107: is this redundant because you already have logging for S3 events that should be skipped?

src/lambda_function/s3_to_glue/app.py
```diff
@@ -31,38 +73,56 @@ def submit_s3_to_json_workflow(messages: list[dict[str,str]], workflow_name):
...
-                workflow_name = os.environ["PRIMARY_WORKFLOW_NAME"]
-        )
+                f"{os.environ['PRIMARY_WORKFLOW_NAME']}: {json.dumps(s3_objects_info)}"
+            )
+            submit_s3_to_json_workflow(
+                objects_info=s3_objects_info,
+                workflow_name=os.environ["PRIMARY_WORKFLOW_NAME"],
+            )
+        # Delete the messages from the queue after all events are successfully submitted
+        for record in event["Records"]:
```
thomasyu888, line 120: What is your rationale for doing this loop at the end and not deleting events as we go?

src/lambda_function/s3_to_glue/template.yaml
```diff
@@ -25,6 +29,19 @@ Parameters:
...
+    Description: >-
+      The number of S3 messages in a SQS event before lambda processes
+      them in a batch
+
+  LambdaMaximumBatchingWindowInSeconds:
+    Type: Number
+    Default: 300
+    Description: >-
+      The amount of time in seconds before lambda polls for all the messages
+      in the SQS queue and processes them in a batch
```
thomasyu888, line 44: Did we want to add a description for why these are the defaults?

thomasyu888 (commented):
🔥 Awesome to see this coming together, just some comments.

src/lambda_function/s3_to_glue/app.py
rxu17: https://docs.aws.amazon.com/lambda/latest/dg/python-context.html
We don't use it but it has to be present in any lambda_handler function. I can add a quick note about it in the function.

src/lambda_function/s3_to_glue/app.py
rxu17: It logs a message about the event and mentions no messages exist. I haven't encountered a scenario where `Records` wouldn't exist, so it would be interesting to see what that looks like. I wonder if such an event would even have a receipt handle for us to delete it from the queue.

src/lambda_function/s3_to_glue/app.py
rxu17: Oh yes, good catch

src/lambda_function/s3_to_glue/app.py
rxu17: Because we call `submit_s3_to_json_workflow` after we get through all of the messages. I don't think we should delete any events until we reach the end where we decide which ones to submit to our S3 to JSON workflow otherwise we would lost that event information if we deleted it earlier and they never make it to the submit workflow step because there was an error with another event.

src/lambda_function/s3_to_glue/app.py
rxu17: Agreed, I wonder if we should create an exception class/tag for just custom exceptions logged by us. Something like this could be NoSourceKeyException, and we define that somewhere. Or, we just mark all of our custom logging with SQSToLambdaWarning: <some_message> so we could easily do a search for these.

src/lambda_function/s3_to_glue/app.py
philerooski: I think it's okay. It's easier to search the logs for "Skipping" rather than constantly have to worry if the messages logged in `filter_object_info` use the same verbiage.

tests/test_s3_to_glue_lambda.py
```diff
@@ -1,70 +1,291 @@
...
         app.submit_s3_to_json_workflow(
-                messages=test_messages,
-                workflow_name="example-workflow"
+            objects_info=test_object_info, workflow_name="example-workflow"
         )
+
+    @mock_sqs
+    def test_that_lambda_handler_does_not_call_submit_s3_to_json_workflow_if_empty_sqs_event(
+        self, empty_queue, test_sqs_queue_url, test_set_env_var
+    ):
```
philerooski, line 143: This isn't something we have to test because it's not part of our application logic. This is something AWS should be worried about.

philerooski (changes requested):
When you reverted your commit, you also lost a bunch of changes which you made to the root README. Was this intentional?

src/lambda_function/s3_to_glue/app.py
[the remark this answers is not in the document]
rxu17: No, just completeness sake/just in case.

src/lambda_function/s3_to_glue/app.py
```diff
@@ -31,38 +73,56 @@ def submit_s3_to_json_workflow(messages: list[dict[str,str]], workflow_name):
...
+    # Initialize SQS client
+    sqs = boto3.client("sqs")
+    s3_objects_info = []
+    if "Records" in event:
+        for record in event["Records"]:
+            s3_event_records = json.loads(record["body"])
+            if "Records" in s3_event_records:
+                for s3_event in s3_event_records["Records"]:
+                    bucket_name = s3_event["s3"]["bucket"]["name"]
+                    object_key = s3_event["s3"]["object"]["key"]
```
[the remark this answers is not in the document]
rxu17: I guess my question is I don't know, but since we should only be allowing S3 events into the SQS queue due to our bucket configuration, it seems they will always be supposedly guaranteed to have this structure. I also don't know if the S3 event will always follow this structure of being under `body` and will contain `Records`. I guess the question here is do we prefer to have a exception to the expected structure run into the `KeyError` or only log it? If we just log it, we may not be as likely to catch it if everything shows up as passed (and it just gets deleted from lambda since it gets processed "successfully")

tests/test_s3_to_glue_lambda.py
rxu17: I think this is just to verify the if no `Records' key in our sqs event message logic in our code unless you are suggesting that we remove that code as well?

src/lambda_function/s3_to_glue/app.py
thomasyu888: Ya, I recommend having json structures, because there are technologies that can parse those out into tabular format.  I'll create a Jira ticket to track

src/lambda_function/s3_to_glue/app.py
thomasyu888: Did you want to add a logger?

src/lambda_function/s3_to_glue/app.py
thomasyu888: Can we add the rationale as a comment.  It is helpful to tell the readers what the code is doing, but we should be able to do that by reading the code.

src/lambda_function/s3_to_glue/app.py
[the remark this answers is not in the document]
philerooski: If anything other than an SQS/S3 event is sent to our lambda that would be a fluke. Meaning: impossible without a human unintentionally submitting a malformed event to the lambda. I think it makes sense to assume that the event will be well formatted (that's for AWS to test and be concerned about, after all). If somehow a message with another format makes its way to our lambda (during testing), it makes sense for the lambda to raise an exception. The message would eventually end up in the DLQ where we can purge it. This scenario occurring in production just isn't feasible unless AWS is broken and we don't need to be prepared for that.

tests/test_s3_to_glue_lambda.py
philerooski: Yes, we can assume that the message is well-formatted.

tests/test_s3_to_glue_lambda.py
```diff
@@ -1,70 +1,291 @@
...
+                patch_client.return_value.receive_message.return_value = empty_queue
+                result = app.lambda_handler(event=empty_queue, context=None)
+                assert result == {"statusCode": 0, "body": "No messages exist"}
+                patch_submit.assert_not_called()
+                patch_client.return_value.assert_not_called()
+
+    @mock_sqs
+    def test_that_lambda_handler_does_not_call_submit_s3_to_json_workflow_if_no_s3_records(
+        self, no_s3_event_records, test_sqs_queue_url, test_set_env_var
+    ):
```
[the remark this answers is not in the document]
philerooski: Yep. See previous comments.

src/lambda_function/s3_to_glue/app.py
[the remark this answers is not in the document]
rxu17: Since we are going with "assume well formatted", then I don't think we need to test that these keys will always be defined in the event since they are supposed to be a part of the S3 events we get.

src/lambda_function/s3_to_glue/app.py
philerooski: I think it's premature to format log messages as JSON before we have a schema for our log messages. If the idea is to have these log messages be consumed by something like snowflake we ought to understand what information we want to include in every log message.

config/prod/namespaced/s3-to-glue-lambda-role.yaml
```diff
@@ -0,0 +1,7 @@
+template:
```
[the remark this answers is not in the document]
rxu17: The only issue is that the lambda role here needs the sqs queue as a dependency and the sqs-queue is namespaced. This is because inside the lambda role's cloudformation, it only gives polling permissions to the namespaced sqs queue. So i think this is why we would need to keep it namespaced

src/lambda_function/s3_to_glue/app.py
rxu17: No, just wanted something for the tests to catch

src/lambda_function/s3_to_glue/app.py
rxu17: So lambda automatically deletes messages after processing them so we don't need these lines of code anymore.

config/prod/namespaced/s3-to-glue-lambda-role.yaml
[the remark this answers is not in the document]
philerooski: I see what you mean. Let's keep it namespaced, then.

src/lambda_function/s3_to_glue/app.py
rxu17: Based on comments below with Phil, it seems like we will just remove the checks for the specific keys (e.g: `Records` in event) in the event JSON because that is the expected format that AWS will provide the SQS message and S3 events in it as and any issues, we'd just get it sent to our DLQ. So probably won't need the logging here.

src/lambda_function/s3_to_glue/app.py
thomasyu888: Going off of what you had suggested in the past, if we logged each record processed, we can potentially have a comparison of records in parquet vs records in JSON.  Agreed that determining a schema for log messages is important.

tests/test_s3_to_glue_lambda.py
rxu17: Hmmm I just realized this. The reason I checked for `Records` key in `s3_event_records` and added this test because I ran into an `KeyError` before when testing the lambda after dropping s3 files indicating there were some things in the SQS message that seemed to be related to S3 events but not contain records..

rxu17: For some reason there's a S3 test event that gets thrown into the bag: 
```
{'Service': 'Amazon S3', 'Event': 's3:TestEvent', 'Time': '2023-09-21T19:43:08.535Z', 'Bucket': 'recover-dev-input-data', 'RequestId': '9YF3QTRANHQ28EY0', 'HostId': 'CXkCznSHquU8oTyCXMJW/gQsatZVAjbvs0YzKxtWYxJD6+QbpZym0az8nW8xJvT3zzRHUt3IuQ0='}
```

philerooski: https://www.mikulskibartosz.name/what-is-s3-test-event

thomasyu888 (approved):
🤖 This is great!  I'm going to pre-approve and count on @philerooski for the final approval :)

philerooski (approved):
One minor fix, but this appears to be working well!

PR #79, opened 2023-09-14
[ETL-538] Add dead letter queue for our SQS queue

philerooski:
The behavior is as follows:

1. If Lambda raises an exception, the message will be sent back to the primary SQS queue up to 3 times before sending that message to the dead letter queue
2. Once the message is in the DLQ, it will be available for up to 2 weeks (the maximum allowed value) until we do a manual redrive of the message back to the primary SQS queue. If the message hasn't been processed after 2 weeks it disappears and will need to be added back to the primary queue either manually (more likely we would generate our own SQS event and send it directly to the Lambda) or by depositing the data in S3 again.

The DLQ doesn't need its own queue policy. It has permissions to receive and redrive messages to the primary queue on account of the primary queue's `RedrivePolicy`.


rxu17 (approved):
LGTM!

PR #80, opened 2023-09-19
[ETL-550] Upload and delete part files immediately after final write

philerooski:
* Splits `write_file_to_json_dataset` into two functions:
  - `write_file_to_json_dataset` contains the logic for writing the JSON part file locally
  - `_upload_file_to_json_dataset` contains the logic for uploading the JSON part file to S3 and acts as a helper function for `write_file_to_json_dataset`.
* `_upload_file_to_json_dataset` handles the upload and potential deletion of the JSON part file, and is called whenever the current part file being written to exceeds the allowed file size (default ~100 MB) or after we have exhausted all data from the input JSON file and must upload the final part file.

### DIFF
--- a/src/glue/jobs/s3_to_json.py
+++ b/src/glue/jobs/s3_to_json.py
@@ -350,7 +350,6 @@ def write_file_to_json_dataset(
     Returns:
         list: A list of files uploaded to S3
     """
-    s3_client = boto3.client("s3")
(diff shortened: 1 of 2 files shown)


thomasyu888 (approved):
🔥 LGTM!  Thanks for doing this.

rxu17 (approved):
LGTM!

PR #81, opened 2023-09-27
[ETL-561] Unquote S3 key name before submitting to Glue

philerooski:
See ticket for details.

### DIFF
--- a/src/lambda_function/s3_to_glue/app.py
+++ b/src/lambda_function/s3_to_glue/app.py
@@ -8,6 +8,7 @@ import os
 import json
 import logging
 import boto3
+from urllib import parse
 
 logger = logging.getLogger()
 logger.setLevel(logging.INFO)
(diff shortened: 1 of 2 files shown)


thomasyu888 (approved):
👍🏼 LGTM!

src/lambda_function/s3_to_glue/app.py
```diff
@@ -114,7 +115,7 @@ def lambda_handler(event, context) -> None:
         else:
             for s3_event in s3_event_records["Records"]:
                 bucket_name = s3_event["s3"]["bucket"]["name"]
-                object_key = s3_event["s3"]["object"]["key"]
+                object_key = parse.unquote(s3_event["s3"]["object"]["key"])
```
rxu17, line 118: Should we have some test cases (or some actual pilot data with the special characters in their name) to test this?

rxu17 (approved):
LGTM, just a comment!

src/lambda_function/s3_to_glue/app.py
thomasyu888: +1 on test case

philerooski: No, that would be testing the behavior of urllib.parse.unquote, which we can assume behaves well. If we look at the behavior of this function:

> Replace %xx escapes with their single-character equivalent.

there aren't any realistic side-effects of using this function vs passing the key as we receive it from SQS. (Unless Care Evolution changed the format of their exports to include unicode encodings? Do we really have to consider that?)

Additionally, this will be tested as part of our integration tests once we complete https://sagebionetworks.jira.com/browse/ETL-562

rxu17: I think it's not necessarily testing `urllib.parse.unquote` directly but rather testing that when we run this `lambda_handler` function entirely, we are submitting object keys to the glue workflow that are expected (in this particular scenario: decoded). This way in the future if we add more code to this `lambda_handler` function, we should still expect that whatever that code is, we will still get our decoded object keys. 

Now whether that is through our CI/CD integration tests with test data and/or through tests in our `test_s3_to_glue_lambda` here, that's fine on either end. It feels like though, if something were to break, it may be worthwhile catching it in the `test_s3_to_glue_lambda` rather than later in our integration tests?

Maybe that warrants a bigger discussion of if we want to enforce an expected format for the names of the S3 files before it gets processed so we don't have a lot of variability in the names and have to handle it each time it pops up.

philerooski: Okay. I refactored app.py and added a test to see if we are handling quoted key names appropriately.

thomasyu888 (approved):

PR #83, opened 2023-10-12
[ETL-558] Decouple S3 to JSON workflow from JSON to Parquet workflow

philerooski:
The important changes happen in [templates/glue-workflow.j2](https://github.com/Sage-Bionetworks/recover/compare/main...etl-568#diff-1ebab210688d33c708f526cb255445c7481a82ab71d012279d4b362c5c712963). This PR splits our workflow into two workflows:

1. S3 to JSON, which is triggered on-demand by the lambda and can run with infinite concurrency.
2. JSON to Parquet, which has a scheduled trigger (default 2 AM UTC / 7 PM PT) and is limited to 1 run at a time. The trigger is only activated if we are deploying to the `main` namespace (so as to not waste resources triggering the pipeline every day on our development stacks).

### DIFF
--- a/templates/glue-workflow.j2
+++ b/templates/glue-workflow.j2
@@ -1,15 +1,21 @@
 AWSTemplateFormatVersion: '2010-09-09'
 
 Description: >-
-    The primary workflow for processing RECOVER data. An outline of the workflow is below:
+    The two workflows for processing RECOVER data. An outline of each workflow is below:
 
-    S3 to JSON ->
+    (S3 to JSON) On-demand trigger (triggered by Lambda) ->
+    (S3 to JSON) S3 to JSON
+
+    and
+
+    (JSON to Parquet) Scheduled trigger ->
+    (JSON to Parquet) Crawler ->
     (JSON to Parquet) EnrolledParticipants and SymptomLog ->
     (JSON to Parquet) HealthKit ->
     (JSON to Parquet) Fitbit ->
     (JSON to Parquet) Google ->
     (JSON to Parquet) Garmin ->
-    CompareParquetJob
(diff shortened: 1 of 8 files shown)


thomasyu888 (approved):
🔥 LGTM here!  Thanks for being intentional about not having the development pipelines run on a daily basis.

templates/glue-workflow.j2
```diff
@@ -72,46 +87,65 @@ Resources:
...
-          LogicalOperator: EQUALS
+      Description: This trigger starts the JSON to Parquet workflow.
+      Type: SCHEDULED
+      Schedule: !Ref JsontoParquetTriggerSchedule
       StartOnCreation: true
-      WorkflowName: !Ref PrimaryWorkflow
+      WorkflowName: !Ref JsonToParquetWorkflow
+
+  JsontoParquetTrigger:
+    Condition: !Not IsMainNamespace
```
thomasyu888, line 138: Weird... it's saying: `Every Condition member must be a string.`. Is that due to this?

BryanFauble: Based on the error log this might help root out the issue and let us test as changes are made:
`aws cloudformation validate-template --template-body templates/glue-workflow.j2`

philerooski: Yeah. CFN is not very flexible. I fixed it.

philerooski: @BryanFauble It seems like that command only checks if the template is valid YAML and not whether it's semantically correct? In any case, I use Jinja in the template so that command fails immediately.

BryanFauble (approved):
With all pipelines passing the changes LGTM!

PR #85, opened 2023-10-19
[ETL-553] Allow merging of notification configs

BryanFauble:
Moved from: https://github.com/Sage-Bionetworks/recover/pull/84

**Problem:**

1. When running lamdba_function `s3_event_config` new notifications could not be added to an S3 bucket, instead the notification configuration was always overwritten.
2. When deleting a NotificationConfiguration for a specific type ALL notification types were deleted.

**Solution:**
A) Adding notifications:
            1) If a bucket has no `NotificationConfiguration` then create the config
            2) If a bucket has a `NotificationConfiguration` but no matching "{destination_type}Configurations" then merge and add the config
            3) If a bucket has a `NotificationConfiguration` and a matching "{destination_type}Configurations":
                3a) If the config is the same then do nothing - ordering of the dict does not matter
                3b) If the config is different then overwrite the matching "{destination_type}Configurations"
B) Deleting notifications - Only deleting the specific notificationConfiguration that matches the `"{destination_type}Configurations"`

**Testing:**
I wrote unit tests for all paths

### DIFF
--- a/src/lambda_function/s3_event_config/app.py
+++ b/src/lambda_function/s3_event_config/app.py
@@ -55,10 +55,14 @@ def add_notification(
 
         Use cases:
             1) If a bucket has no `NotificationConfiguration` then create the config
-            2) If a bucket has a `NotificationConfiguration` but no matching "{destination_type}Configurations" then merge and add the config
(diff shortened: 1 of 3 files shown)


BryanFauble:
Closing this PR - Will reopen with changes. What this does and what we want do not line up.

PR #86, opened 2023-10-20
[ETL-567] Use ECS logging in S3 to JSON job

philerooski:
I've added comments to annotate my changes.


src/glue/jobs/s3_to_json.py
```diff
@@ -402,6 +447,22 @@ def write_file_to_json_dataset(
...
+                    "event.category": ["file"],
+                    "event.type": ["info", "creation"],
+                    "event.action": "list-file-properties",
+                    "labels": {
+                        k: v.isoformat()
+                        if isinstance(v, datetime.datetime) else v
+                        for k, v in metadata.items()
+                    }
+                }
+    )
```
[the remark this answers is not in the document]
thomasyu888: Should these comments be actual comments in the code?

src/glue/jobs/s3_to_json.py
```diff
@@ -523,9 +598,30 @@ def get_metadata(basename: str) -> dict:
         and basename_components[-2] == "Deleted"
     ):
         metadata["type"] = "{}_Deleted".format(metadata["type"])
-    logger.debug("metadata = %s", metadata)
     return metadata
 
+def get_basic_file_info(file_path):
```
thomasyu888, line 603: Nit: add type hints

src/glue/jobs/s3_to_json.py
```diff
@@ -84,12 +86,30 @@ def transform_object_to_array_of_objects(
...
             key_value = key_type(k)
         except ValueError:
-            logger.warning(value_error, k,key_type)
+            logger.warning(
+                    value_error, k, key_type,
+                    extra={
+                        **logger_context,
+                        "event.kind": "alert",
+                        "event.category": ["configuration"],
+                        "event.type": ["change"],
```
BryanFauble, line 95: Thoughts on the `deletion` event.type?
https://www.elastic.co/guide/en/ecs/current/ecs-allowed-values-event-type.html#ecs-event-type-deletion
vs
https://www.elastic.co/guide/en/ecs/current/ecs-allowed-values-event-type.html#ecs-event-type-change

I'm leaning towards `deletion` myself because we are destroying the `key_value` in this case rather than a transformation into some other useful value.

src/glue/jobs/s3_to_json.py
```diff
@@ -164,9 +187,19 @@ def transform_json(
...
+                        logger.error(
+                                (f"Problem CustomFields.{field_name}: "
+                                f"{json_obj['CustomFields'][field_name]}"),
+                                extra={
+                                    **logger_context,
+                                    "error.message": repr(e),
+                                    "error.type": "json.JSONDecodeError",
+                                    "event.kind": "alert",
+                                    "event.category": ["configuration"],
+                                    "event.type": ["creation"],
```
BryanFauble, line 199: Same thought as the last comment - thinking about an event type of `deletion`.

src/glue/jobs/s3_to_json.py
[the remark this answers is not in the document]
BryanFauble: +1

src/glue/jobs/s3_to_json.py
```diff
@@ -552,20 +648,47 @@ def process_record(
...
+            with z.open(json_path, "r") as f:
+                line_count = sum(1 for _ in f)
+            basic_file_info = get_basic_file_info(file_path=json_path)
+            logger.debug(
+                    "Getting metadata",
+                    extra={
+                        **basic_file_info,
+                        "event.kind": "event",
+                        "event.category": ["process"],
+                        "event.type": ["start"]
```
BryanFauble, line 661: If we have a `start` event.type - should we have a a respective end after it's been finished?

https://www.elastic.co/guide/en/ecs/current/ecs-allowed-values-event-type.html#ecs-event-type-end

src/glue/jobs/s3_to_json.py
```diff
@@ -552,20 +648,47 @@ def process_record(
...
-            dataset_identifier = metadata["type"]
-            logger.info("Writing %s to dataset %s",
-                        json_path, dataset_identifier)
+            metadata_str_keys = {
+                k: v.isoformat()
+                if isinstance(v, datetime.datetime) else v
+                for k, v in metadata.items()
+            }
+
+            logger_context = {**basic_file_info, "labels": metadata_str_keys}
```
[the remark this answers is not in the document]
BryanFauble: +1 I really like your idea here to propagate along context into all of the log messages.

thomasyu888 (commented):
🔥 Nice!  Just some comments.

src/glue/jobs/s3_to_json.py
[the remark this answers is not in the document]
philerooski: Maybe? I was taught a lightweight commenting philosophy: Use the docstring to describe the behavior of code and only use inline comments when there is particularly confusing logic. Otherwise, let the code speak for itself.

In this specific case, is it helpful to have an inline comment saying that this log message includes the line count when the line:
```
 "file.LineCount": line_count
```
is there in plain sight? I think making note of the line count log makes more sense in the docstring.

src/glue/jobs/s3_to_json.py
philerooski: I see what you mean. Deletion does make more sense here.

src/glue/jobs/s3_to_json.py
philerooski: In general, perhaps. I actually think we should remove this log message altogether. This was one of the log messages that was originally just plaintext and I tried to convert it to ECS.  It made sense when I was just beginning to develop the script and seeing where things were breaking, but getting metadata is a very lightweight process. Adding 18 extra lines of logging code to say (begin parsing metadata from the file name) and (end parsing metadata from the filename) seems a bit extreme.

src/glue/jobs/s3_to_json.py
philerooski: @BryanFauble I'm having trouble differentiating how we would log two very different types of events. For example:

1. Failing to delete a configuration setting or parameter:
```
logger.error(
    "Failed to delete configuration or parameter",
    extra={
        "event.kind": "alert",
        "event.category": ["configuration"],
        "event.type": ["deletion"],
        "event.outcome": "failure"
    }
)                      
```

2. Failing to do some operation on data, resulting in that data being dropped/deleted:
```
logger.error(
    "Failed to do an operation on data. We will drop/delete this data instead.",
    extra={
        "event.kind": "alert",
        "event.category": ["configuration"],
        "event.type": ["deletion"],
        "event.outcome": "??????"
    }
)
# Code to drop data goes here           
```
I'm starting to think that `"event.type": ["change"]` and `"event.outcome": "failure"` makes more sense here.

BryanFauble: I am thinking of these log messages as "What is the result of the thing I tried to do".
For your example what do you think of:
```
try:
    # Do something
except:
    logger.error(
        "Failed to do an operation on data. We will drop/delete this data instead.",
        extra={
            "event.kind": "alert",
            "event.category": ["configuration"],
            "event.type": ["deletion"],
            "event.outcome": "failure"
        }
    )
    try:
        # Do the delete
    except:
        logger.error(
            "Failed to delete configuration or parameter",
            extra={
                "event.kind": "alert",
                "event.category": ["configuration"],
                "event.type": ["error"],
                "event.outcome": "failure"
            }
        )    
```

philerooski: Taking a hint from the `event.outcome` ECS docs:
> Note that when a single transaction is described in multiple events, each event may populate different values of event.outcome, according to their perspective.

I think this makes more sense:
```
try:
    # Do something
except:
    logger.error(
        "Failed to do an operation on data",
        extra={
            "event.kind": "alert",
            "event.category": ["configuration"],
            "event.type": ["change"],
            "event.outcome": "failure"
        }
    )
    try:
        # Do the delete
        logger.warning(
            "Deleted some data",
            extra={
                "event.kind": "alert",
                "event.category": ["configuration"],
                "event.type": ["deletion"],
                "event.outcome": "success"
            }
        )
    except:
        logger.error(
            "Failed to delete data",
            extra={
                "event.kind": "alert",
                "event.category": ["configuration"],
                "event.type": ["deletion"],
                "event.outcome": "failure"
            }
        )  
```

src/glue/jobs/s3_to_json.py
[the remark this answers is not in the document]
BryanFauble: This function is cognitively complex (Branching paths, nested for loops, and some setup before it gets there). The docstring does a great job to explain what the function is doing, but in-line comments give a great how perspective.
```

                for transformed_record in transformed_block:
                    # Sum total line count of the file by counting lines in each individual block
                    line_count += 1
```

On a side note after reading a lot of code I'm trying to push towards the "Never Nesting" methodology personally where I can. https://www.youtube.com/watch?v=CFRhGnuXG-4 is a great video I liked on the topic

src/glue/jobs/s3_to_json.py
BryanFauble: I like it

src/glue/jobs/s3_to_json.py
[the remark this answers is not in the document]
philerooski: > On a side note after reading a lot of code I'm trying to push towards the "Never Nesting" methodology

I like this! I often have this complexity problem.

I'm going to refactor some of the code now. Maybe I can get it to the point where in-line comments are effectively redundant with the function calls but if not I can add inline comments.

philerooski:
@BryanFauble @thomasyu888 I refactored `s3_to_json.py` to unnest some of the logic and make things more readable/testable. I ran the same data through the existing (dev/main) pipeline and compared the MD5 of the resulting JSON files in the JSON datasets to make sure that nothing changed in the way we are processing the data. I think it will be less confusing if you compare changes made in [just that refactoring commit](https://github.com/Sage-Bionetworks/recover/pull/86/commits/8f15a6e19ac6a2c21d783323f99db9e79a7eeaec).

src/glue/jobs/s3_to_json.py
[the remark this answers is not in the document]
philerooski: I was able to pull out some of the logic into its own function, but most of the code has to do with setup/initialization and so I've added inline comments to point out what each section is doing.

src/glue/jobs/s3_to_json.py
```diff
@@ -81,51 +81,78 @@ def transform_object_to_array_of_objects(
...
+                    logger_context=logger_context
+            )
         obj = {
             key_name: key_value,
             value_name: value_value
         }
         array_of_obj.append(obj)
     return array_of_obj
 
+def _log_error_transform_object_to_array_of_objects(
```
thomasyu888, line 112: +1 to this helper function, by refactoring, the code also becomes "easier" to write unit tests for.

thomasyu888 (approved):

BryanFauble (approved):
LGTM!

src/glue/jobs/s3_to_json.py
```diff
@@ -235,42 +231,158 @@ def transform_json(
...
+                                **logger_context,
+                                "error.message": repr(error),
+                                "error.type": "json.JSONDecodeError",
+                                "event.kind": "alert",
+                                "event.category": ["change"],
+                                "event.type": ["error"],
+                                "event.outcome": "failure",
+                            }
+                    )
+                    json_obj["CustomFields"][field_name] = []
```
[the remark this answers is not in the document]
philerooski: That's important to consider. Probably the correct order looks something like this:

```
try:
  do_something()
except:
  log_failed_to_do_something()
  value = None # fallback action
  log_did_fallback_action()
```

or maybe:

```
try:
  do_something()
except:
  log_failed_to_do_something()
  log_attempting_fallback_action()
  value = None # fallback action
```

This way we can capture that we failed to `do_something` before attempting any fallback action (which might also fail). In this case, the fallback action is so simple that the ordering isn't as critical.

PR #87, opened 2023-10-20
[ETL-553] Allow merging of notification configs

BryanFauble:
**Problem:**

1. When running lamdba_function `s3_event_config` new notifications could not be added to an S3 bucket, instead the notification configuration was always overwritten.
2. When deleting a NotificationConfiguration for a specific type ALL notification types were deleted.

**Solution:**
A) Adding notifications:
            1) If a bucket has no `NotificationConfiguration` then create the config
            2) If a bucket has a `NotificationConfiguration` but no matching
                "{destination_arn}" for the "{destination_type}" then add the config
            3) If a bucket has a `NotificationConfiguration` and a matching
                                    "{destination_arn}" for the "{destination_type}":
                3a) If the config is the same then do nothing
                3b) If the config is different then overwrite the matching config
B) Deleting notifications - Only deleting the specific notificationConfiguration that matches the `"{destination_type}Configurations"` and destination arn

**Testing:**
I wrote unit tests for all paths

### DIFF
--- a/src/lambda_function/s3_event_config/app.py
+++ b/src/lambda_function/s3_event_config/app.py
@@ -4,11 +4,14 @@ puts a s3 event notification configuration for the S3 to Glue lambda in the
 input data S3 bucket set by the environment variable `S3_SOURCE_BUCKET_NAME`.
 
 This Lambda app also has the option of deleting the notification configuration
-from an S3 bucket
+from an S3 bucket.
+
+Only certain notification configurations work. Only `QueueConfigurations` are expected.
 """
 import os
 import json
 import logging
+import typing
 
 import boto3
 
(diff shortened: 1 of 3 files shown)


src/lambda_function/s3_event_config/app.py
```diff
@@ -56,35 +161,117 @@ def add_notification(
...
+    (
+        existing_bucket_notification_configuration,
+        existing_notification_configurations_for_type,
+    ) = get_existing_bucket_notification_configuration_and_type(
+        s3_client, bucket, destination_type
+    )
+
+    (
+        index_of_matching_arn,
+        matching_notification_confiugration,
```
philerooski, line 246: Not that it matters, but there's a typo here. What do you think of using `_` instead to indicate that we don't use this variable later in the code?

src/lambda_function/s3_event_config/app.py
```diff
@@ -40,14 +48,111 @@ def lambda_handler(event, context):
         raise KeyError(err_msg)
 
 
+def get_existing_bucket_notification_configuration_and_type(
+    s3_client: boto3.client, bucket: str, destination_type: str
+) -> typing.Tuple[dict, list]:
```
philerooski, line 53: This would be more readable as a named tuple or dict.

src/lambda_function/s3_event_config/app.py
```diff
@@ -56,35 +161,117 @@ def add_notification(
...
+        existing_notification_configurations_for_type,
+        destination_arn,
+    )
 
-def delete_notification(s3_client: boto3.client, bucket: str):
-    """Deletes the S3 notification configuration from an existing bucket
+    if index_of_matching_arn is not None:
+        if json.dumps(
+            matching_notification_configuration, sort_keys=True
+        ) != json.dumps(new_notification_configuration, sort_keys=True):
```
philerooski, line 195: I still think it would be cleaner to compare the "index" of these two notifications configurations, the index being the necessary properties `Events` and `Filter.Key.FilterRules.Value` in our notification configuration. I think it's less likely that these fields will change in the future versus some arbitrary change in the response. But this works for now 👍

src/lambda_function/s3_event_config/app.py
```diff
@@ -40,14 +48,111 @@ def lambda_handler(event, context):
...
+        existing_bucket_notification_configuration,
+        existing_notification_configurations_for_type,
+    )
+
+
+def get_matching_notification_configuration(
+    destination_type_arn: str,
+    existing_notification_configurations_for_type: list,
+    destination_arn: str,
+) -> typing.Union[tuple[int, dict], tuple[None, None]]:
```
philerooski, line 97: Likewise, this would be more readable as a named tuple or dict.

philerooski (approved):
Awesome. I left some comments but I'm approving this as is.

I realized we've created a new problem for ourselves by accumulating bucket event notification rules. This is a super minor issue, but I've filed a ticket here to handle that a bit more elegantly. Maybe we can piggyback these changes on some other work that needs to be done: https://sagebionetworks.jira.com/browse/ETL-572

It's also worth considering why this lambda ought to be namespaced at all. Why not make it a general use lambda that is invoked by all our branches? The current implementation is perfectly functional so I don't think this is a particularly high priority change, either.

thomasyu888 (approved):
🔥 LGTM!

src/lambda_function/s3_event_config/app.py
BryanFauble: I like this thought. I will extract this out to a `compare` function so all the fields we are checking is in one spot.

BryanFauble: I added this to: https://github.com/Sage-Bionetworks/recover/pull/87/commits/b473c56fcd1a4c4d35131fcf366c740e0564fee7

src/lambda_function/s3_event_config/app.py
BryanFauble: I added this to: https://github.com/Sage-Bionetworks/recover/pull/87/commits/b473c56fcd1a4c4d35131fcf366c740e0564fee7 - Is this what you had in mind?

src/lambda_function/s3_event_config/app.py
BryanFauble: I added this in https://github.com/Sage-Bionetworks/recover/pull/87/commits/f7c2e3a5613aae43c90a9cd0300e1995aa9b96e6

src/lambda_function/s3_event_config/app.py
philerooski: This works! I oftentimes use dict to return multiple results in a clear way but the way typing works with a class is pretty nice.

philerooski:
@BryanFauble feel free to merge whenever you're comfortable with the changes. The PR is already approved.

PR #88, opened 2023-10-30
[ETL-575] Pass SkipDestinationValidation=True

BryanFauble:
In the lambda function to add event notifications pass `SkipDestinationValidation=True` in the `put_bucket_notification_configuration` call


philerooski (approved):
Your log messages are f-strings but you don't use the f-string syntax.

Do you need to add a space after "for" or does `create_formatted_message` include white space at the beginning of its message?

PR #89, opened 2023-11-02
[ETL-568] Script to compare read/write record count from S3 to JSON logs

philerooski:
Changes:

* `src/glue/jobs/s3_to_json.py`
  - Disabled propagation of LogRecords to root logger so that log messages contain _only_ the JSON formatted log emitted by the ECS log handler. This enables us to use Cloudwatch Metrics Insights to query our JSON.
  - Updated tests since there is a pytest bug where the caplog fixture only captures logs emitted by the root logger.
  - Included the workflow run ID as the `process.parent.pid` in the logger context so that every log message can be associated with a specific workflow run.

* `src/scripts/consume_logs.py`
  - Initial commit
  - Has tests
  - The first half (see `main`) is general functionality that will run an arbitrary query over a timeframe and log group that you specify and group those logs by their workflow run. The second half is functionality specific to comparing the line counts of file read/write. We can repurpose the general functionality later if we ever have more tests that run against the logging data.
  - A "comparison report" is written to CSV which contains the read/write line counts of every file processed and some contextual information (workflow run ID and cohort).
  - If any differences between read/write count are discovered, a separate "missing data" CSV report is written. It's formatted the same as the comparison report but only contains files where a difference was discovered.

### DIFF
--- a/src/scripts/consume_logs/consume_logs.py
+++ b/src/scripts/consume_logs/consume_logs.py
@@ -1,0 +1,333 @@
+"""
+This script will compare read/write line counts of files processed by the
+S3 to JSON job by querying the Cloudwatch log group where that job writes
+its logs.
+
+A message is printed whether a mismatch between read/write counts was discovered,
+indicating that data went missing or, potentially, that additional data not present
+in the original JSON file was written to the JSON dataset.
+
+A CSV file is written (consume_logs_comparison_report.csv) to the current directory
+containing the read/write counts for every file matched in the query. This CSV file
+contains the fields:
+
+    * workflow_run_id : the ID of the workflow run
+    * cohort : the cohort this data belongs to (adults_v1, pediatric_v1, etc.)
+    * file_name : the file name.
+    * line_count_access : the number of lines counted during read of the original JSON file.
+    * line_count_creation : the number of lines counted during write to the JSON dataset.
+    * line_count_difference : line_count_access - line_count_creation
+
+If missing data is discovered, another CSV file (consume_logs_missing_data_report.csv)
+is written which contains only the subset of files which have a line_count_difference != 0.
+The schema is the same as above.
+"""
+import datetime
+import json
+import time
+from collections import defaultdict
+from typing import List, Dict
+
+import argparse
+import boto3
+import pandas
+
+def read_args():
+    parser = argparse.ArgumentParser(
+            description="Query S3 to JSON CloudWatch logs and compare line counts.",
+            formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    parser.add_argument(
+            "--log-group-name",
+            help="The name of the log group to query.",
+            default="/aws-glue/python-jobs/error"
+    )
+    parser.add_argument(
+            "--query",
+            help="The query to run against the log group.",
+            default='fields @message | filter event.action = "list-file-properties"'
+    )
+    parser.add_argument(
+            "--start-datetime",
+            help=(
+                "Query start time (local time) expressed in a format parseable by "
+                "datetime.datetime.strptime. This argument should be "
+                "formatted as `--time-format`."),
+            required=True,
+    )
+    parser.add_argument(
+            "--end-datetime",
+            help=(
+                "Default is \"now\". Query end time (local time) expressed "
+                "in a format parseable by datetime.datetime.strptime. This "
+                "argument should be formatted as `--time-format`."),
+    )
+    parser.add_argument(
+            "--datetime-format",
+            help="The time format to use with datetime.datetime.strptime",
+            default="%Y-%m-%d %H:%M:%S",
+    )
+    args = parser.parse_args()
+    return args
+
+def get_seconds_since_epoch(datetime_str: str, datetime_format: str) -> int:
+    """
+    Returns the seconds since epoch for a specific datetime (local time)
+
+    Args:
+        datetime_str (str): A datetime represented by a format parseable by
+            datetime.datetime.strptime.
+        datetime_format (str): The datetime.datetime.strptime format used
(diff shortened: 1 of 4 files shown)


src/scripts/consume_logs/consume_logs.py
```diff
@@ -0,0 +1,312 @@
...
+            "line_count_difference != 0"
+        )
+        if len(missing_data != 0):
+            print(
+                    "Discovered differences between records read/write "
+                    f"in workflow run {workflow_run}"
+            )
+            missing_data = missing_data.assign(workflow_run_id = workflow_run)
+            all_missing_data = pandas.concat([all_missing_data, missing_data])
+    if len(all_missing_data):
```
thomasyu888, line 247: Did you mean to have `len(all_missing_data) > 0`

src/scripts/consume_logs/consume_logs.py
```diff
@@ -0,0 +1,312 @@
...
+        event_comparison = access_events.merge(
+                creation_events,
+                how="left",
+                on=["cohort", "file_name"],
+                suffixes=("_access", "_creation")
+        )
+        event_comparison["line_count_difference"] = \
+                event_comparison["line_count_access"] - event_comparison["line_count_creation"]
+        workflow_run_event_comparison[workflow_run] = event_comparison
+    report_results(
```
thomasyu888, line 307: Do you have a csv you can share with me via synapse?

src/scripts/consume_logs/consume_logs.py
```diff
@@ -0,0 +1,312 @@
...
+    )
+    parser.add_argument(
+            "--datetime-format",
+            help="The time format to use with datetime.datetime.strptime",
+            default="%Y-%m-%d %H:%M:%S",
+    )
+    args = parser.parse_args()
+    return args
+
+def get_seconds_since_epoch(datetime_str: str, datetime_format: str):
```
thomasyu888, line 48:
```suggestion
def get_seconds_since_epoch(datetime_str: str, datetime_format: str) -> int:
```

src/scripts/consume_logs/consume_logs.py
thomasyu888, line 1: This is a great showcase, I wonder if instead of the RECOVER data being in snowflake, I ask brad E if we can have this log data being consumed by snowflake so we can have a live dashboard of the processing times as a starting point.

What i'm thinking is a "live" dash for how many records are being processed.

src/scripts/consume_logs/consume_logs.py
```diff
@@ -0,0 +1,312 @@
...
+            endTime=end_unix_time,
+            queryString=query_string
+    )
+    query_response = logs_client.get_query_results(
+            queryId=start_query_response["queryId"]
+    )
+    _check_for_failed_query(query_response)
+    while query_response["status"] != "Complete":
+        time.sleep(1)
+        query_response = logs_client.get_query_results(
```
thomasyu888, line 111: Nit: There's probably some fun stuff you can do with https://docs.python.org/3/library/asyncio.html here. But for a PoC, that may be overkill

src/scripts/consume_logs/consume_logs.py
```diff
@@ -0,0 +1,312 @@
...
+            {field['field']: field['value'] for field in log_message}
+            for log_message in query_results
+    ]
+    workflow_run_logs = defaultdict(list)
+    for log_record in log_records:
+        log_message = json.loads(log_record["@message"])
+        workflow_run_logs[log_message["process"]["parent"]["pid"]].append(log_message)
+    return workflow_run_logs
+
+def transform_logs_to_dataframe(log_messages: List[dict]) -> "pandas.DataFrame":
```
thomasyu888, line 149: Nit: Why string "pandas.DataFrame"?  It _should_ work without it.

src/scripts/consume_logs/consume_logs.py
```diff
@@ -0,0 +1,312 @@
...
+                )
+                or all(
+                    [
+                        k in log_message["event"]["type"]
+                        for k in ["access", "creation"]
+                    ]
+                )
+        ):
+            raise KeyError(
+                    "Did not find event.type in log message or "
```
thomasyu888, line 185: What would cause this?

src/scripts/consume_logs/consume_logs.py
```diff
@@ -0,0 +1,312 @@
...
+    all_comparisons = pandas.concat(
+            workflow_run_event_comparison,
+            names=["workflow_run_id", "index"]
+    )
+    all_missing_data = pandas.DataFrame()
+    for workflow_run in workflow_run_event_comparison:
+        missing_data = workflow_run_event_comparison[workflow_run].query(
+            "line_count_difference != 0"
+        )
+        if len(missing_data != 0):
```
thomasyu888, line 240: Is this if condition wrong?

thomasyu888 (approved):
🤖 I'm going to pre-approve, but I have some minor comments.  Incredible work!

src/scripts/consume_logs/consume_logs.py
philerooski: In general, what will break the assumptions of this script is if we added/updated logs which match the query (`file.action` = `list-file-properties`) but don't conform to the format we expect from such logs. This particular case checks that `event.type` contains "access" xor "creation". This would never happen with the current implementation of S3 to JSON, but changes we make in the future could potentially break the assumptions we make if we aren't careful.

src/scripts/consume_logs/consume_logs.py
philerooski: Typo.

BryanFauble (approved):
No additional comments!

LGTM!

PR #90, opened 2023-11-07
[ETL-562] Copy test data for integration testing

BryanFauble:
Problem:
Previously we were manually creating test events to test the whole pipeline - instead we wanted the entire flow to handle this.

Solution:
Updating the integration test step to do an `aws cp` command to copy the test ingestion data

Testing:
See comment section of https://sagebionetworks.jira.com/browse/ETL-562 to see the whole flow having ran

### DIFF
--- a/.github/workflows/upload-and-deploy.yaml
+++ b/.github/workflows/upload-and-deploy.yaml
@@ -278,6 +278,7 @@ jobs:
       - name: Copies over test files from ingestion bucket
         run: >
           aws s3 cp s3://recover-dev-ingestion/pilot-data/ s3://$DEV_INPUT_BUCKET/$NAMESPACE/
+          --recursive
           --exclude "owner.txt"
 
   sceptre-deploy-staging:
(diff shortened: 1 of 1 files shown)


.github/workflows/upload-and-deploy.yaml
```diff
@@ -272,27 +272,14 @@ jobs:
...
 
-      - name: generate test events
-        run: >
-          pipenv run python src/lambda_function/s3_to_glue/events/generate_test_event.py
-          --input-bucket $DEV_INPUT_BUCKET
-          --input-key-prefix $NAMESPACE
-          --output-directory ./src/lambda_function/s3_to_glue/events/
-
       - name: Setup sam
         uses: aws-actions/setup-sam@v2
```
philerooski, line 276: We don't need this step since we are no longer dealing with the lambda.

philerooski (approved):
One minor change.

.github/workflows/upload-and-deploy.yaml
rxu17, line 1: We could also remove [these lines](https://github.com/Sage-Bionetworks/recover/blob/dfa9ed38317f1ef2757820139e820450fb167a7b/.github/workflows/upload-and-deploy.yaml#L58-L61).

These lines were placed there originally for when we were still generating a set of custom SQS events which we submit directly to the S3 to Glue lambda.

rxu17 (approved):
LGTM! just a minor comment

PR #93, opened 2023-12-08
[ETL-580] JSON to Parquet - Write record count of each export to S3 

philerooski:
Summary of changes:

---

* `src/glue/jobs/json_to_parquet.py`
* `tests/test_json_to_parquet.py`

Whenever a read, write, or transform operation is done with the data, we record the number of records for each `export_end_date` (which acts as an identifier for the original export). To support this functionality, functions which did any reading/writing/transforming were split up to be more atomic. Each of these functions requires a `record_counts` and `logger_context` object, which is passed to the function responsible for doing the counting: `count_records_for_event`.

Since there are typically multiple read/write/transform events per data type, we accumulate these within `record_counts` throughout the job. As the very last step in the job, we concatenate each of these event-specific counts within each data type and write them as a CSV file to S3. Oftentimes there is only one data type which needs counts in a job (after all, the jobs are data type specific), but for data types which have an associated "deleted" table containing deleted samples, we do counts for those under a separate data type. The docstrings for `count_record_for_event` and `store_record_counts` go into the nitty gritty details.

Originally I had tried logging the count information in the ECS format, hence there being some ancillary changes related to ECS logging. This didn't work out (see Jira ticket), but the ECS work will support future work on https://sagebionetworks.jira.com/browse/ETL-573 .

Also, a note on tests: JSON to Parquet tests were originally written as a blend of unit and integration tests. This will eventually need to be rewritten as pure unit tests, so any additional tests added have been written as pure unit tests. Fixing the rest of the tests was outside the scope of this ticket (which already took me ~3 weeks to complete) so those are mostly unchanged.

---

* `tests/test_json_to_parquet/TestFlatInsertedDateDataType_20230512.ndjson`
* `tests/test_json_to_parquet/TestFlatInsertedDateDataType_20230612.ndjson`

We expect our JSON to always contain an `export_end_date` field since we use that field while counting records. This adds that field to these test data.

---

* `config/develop/namespaced/glue-job-JSONToParquet.yaml`
* `config/prod/namespaced/glue-job-JSONToParquet.yaml`
* `templates/glue-job-JSONToParquet.j2`
* `tests/Dockerfile.aws_glue_4`

Add support for additional python modules in JSON to Parquet job (specifically, ECS logging)

---

* `src/glue/jobs/s3_to_json.py`

Added a `merge_dicts` function which resolves the potential issue where: if a dict is passed to the logger, it could overwrite the same dict in the `logger_context`. Merging dictionaries in a smart way is not trivial, but I think this handles the most obvious case pretty well. Specifically: When passing additional `labels` (a dict), we merge this with the `labels` in the `logger_context`. Previously, the locally defined `labels` would overwrite the `labels` in the `logger_context`.

### DIFF
--- a/src/glue/jobs/json_to_parquet.py
+++ b/src/glue/jobs/json_to_parquet.py
@@ -10,17 +10,29 @@ dataset and an `healthkitv2heartbeat_subsamples` dataset.
 
 """
 
+import logging
 import os
 import sys
-from datetime import datetime
+from collections import defaultdict
+from copy import deepcopy
 
 import boto3
+import ecs_logging
+import pandas
 from awsglue import DynamicFrame
 from awsglue.context import GlueContext
 from awsglue.job import Job
 from awsglue.gluetypes import StructType
 from awsglue.utils import getResolvedOptions
 from pyspark import SparkContext
+
+# Configure logger to use ECS formatting
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.DEBUG)
+handler = logging.StreamHandler()
+handler.setFormatter(ecs_logging.StdlibFormatter())
+logger.addHandler(handler)
+logger.propagate = False
 
 INDEX_FIELD_MAP = {
     "enrolledparticipants": ["ParticipantIdentifier"],
@@ -92,36 +104,34 @@ def has_nested_fields(schema: StructType) -> bool:
     Returns:
         bool: Whether this schema contains struct or array fields.
     """
-    for col in schema:
-        if col.dataType.typeName() in ["array", "struct"]:
-            return True
+    array_struct_cols = [
+        col.dataType.typeName() in ["array", "struct"]
+        for col in schema
+    ]
+    if any(array_struct_cols):
+        return True
     return False
 
 def get_table(
         table_name: str,
         database_name: str,
-        glue_context: GlueContext
+        glue_context: GlueContext,
+        record_counts: dict,
+        logger_context: dict,
     ) -> DynamicFrame:
     """
     Return a table as a DynamicFrame with an unambiguous schema.
-    Duplicate samples are dropped by referencing the `InsertedDate`
-    field, keeping the more recent sample. For any data types which
-    don't have this field, we drop duplicates by referencing `export_end_date`.
-    Additionally, we drop any superfluous partition_* fields which are
-    added by Glue.
 
     Args:
         table_name (str): The name of the Glue table.
         database_name (str): The name of the Glue database
         glue_context (GlueContext): The glue context
+        logger_context (dict): A dictionary containing contextual information
+            to include with every log.
 
     Returns:
         awsglue.DynamicFrame: The Glue Table's data as a DynamicFrame
-            after the duplicates have been dropped by descending
-            export date.
     """
-    table_name_components = table_name.split("_")
-    table_data_type = table_name_components[1]
     table = (
             glue_context.create_dynamic_frame.from_catalog(
                  database=database_name,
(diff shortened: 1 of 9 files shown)


BryanFauble (approved):
Excellent work!

src/glue/jobs/s3_to_json.py
```diff
@@ -554,10 +555,10 @@ def write_file_to_json_dataset(
                 delete_upon_successful_upload=delete_upon_successful_upload
         )
         uploaded_files.append(current_output_path)
-        logger.info(
-                "Output file attributes",
-                extra={
-                    **logger_context,
+        logger_extra = dict(
+            merge_dicts(
```
rxu17, line 559: Is this something we should make sure to do for everywhere that we are using logger along with `logger_context `and specifying the `extra` argument in case we have more overwrites of the `logger_context`'s dictionary items in the future? Or put a flag near where logger_context is first created?

rxu17 (approved):
LGTM! Just a few comments

src/glue/jobs/json_to_parquet.py
```diff
@@ -332,13 +414,133 @@ def write_table_to_s3(
             format = "parquet",
             transformation_ctx="write_dynamic_frame")
 
+def count_records_for_event(
+        table: "pyspark.sql.dataframe.DataFrame",
+        event: str,
```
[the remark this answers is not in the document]
philerooski: Technically I can't. StrEnum was introduced in 3.11 and this runs in 3.10.

I've never used StrEnum as a type hint before. From the Enum tutorial it seems like it was designed to provide syntactic sugar for working with classes, rather than as a type hint. Is there a benefit to adding the type hint in addition to the docstring and the check we do inside the function itself?

src/glue/jobs/json_to_parquet.py
```diff
@@ -92,36 +104,34 @@ def has_nested_fields(schema: StructType) -> bool:
...
+        glue_context: GlueContext,
+        record_counts: dict,
+        logger_context: dict,
     ) -> DynamicFrame:
     """
     Return a table as a DynamicFrame with an unambiguous schema.
-    Duplicate samples are dropped by referencing the `InsertedDate`
-    field, keeping the more recent sample. For any data types which
-    don't have this field, we drop duplicates by referencing `export_end_date`.
-    Additionally, we drop any superfluous partition_* fields which are
```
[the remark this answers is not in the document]
philerooski: Nice catch. I forgot to remove that part of the docstring back into the `get_table` function.

The partition_* fields are just an eyesore. We already get our partition fields as regular fields (e.g., `cohort`), but Glue adds a `partition_cohort` field when reading the data as a dynamic frame.

src/glue/jobs/json_to_parquet.py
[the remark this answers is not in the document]
BryanFauble: [I was more thinking of Enum in general](https://docs.python.org/3.10/library/enum.html#creating-an-enum) not specifically a `StrEnum` (I didn't even know that was a thing).

> Member values can be anything: [int](https://docs.python.org/3.10/library/functions.html#int), [str](https://docs.python.org/3.10/library/stdtypes.html#str), etc..

```
class EventType(Enum):
    """The event associated with a count."""
    
    
    READ = "READ"
    """This table has just now been read from the Glue table catalog
                and has not yet had any transformations done to it."""

    DROP_DUPLICATES = "DROP_DUPLICATES"
    """This table has just now had duplicate records dropped
                (see the function `drop_table_duplicates`)."""

    DROP_DELETED_SAMPLES = "DROP_DELETED_SAMPLES"
    """This table has just now had records which are
                present in its respective "Deleted" table dropped (see the function
                `drop_deleted_healthkit_data`)."""

    WRITE = "WRITE"
    """This table has just now been written to S3."""


def do_thing(event_type: EventType):
    print(event_type)


do_thing(EventType.READ)
do_thing(EventType.DROP_DUPLICATES)
do_thing(EventType.DROP_DELETED_SAMPLES)
do_thing(EventType.WRITE)
```

The point around this is that I don't need to know the string constant to use. I could check the docstring, but I could also use one of the available Enum values:

image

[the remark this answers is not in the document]
philerooski: That makes sense. It has some nifty features. I'll try it out.

src/glue/jobs/s3_to_json.py
```diff
@@ -643,6 +646,41 @@ def _upload_file_to_json_dataset(
         os.remove(file_path)
     return s3_output_key
 
+def merge_dicts(x: dict, y: dict) -> Generator:
```
thomasyu888, line 649: Nit: Just out of curiosity, do you know if there is a feature that already exists from some package?

thomasyu888 (approved):
💯 Awesome work!

src/glue/jobs/json_to_parquet.py
```diff
@@ -332,13 +413,141 @@ def write_table_to_s3(
...
             transformation_ctx="write_dynamic_frame")
 
+class CountEventType(Enum):
+    """The event associated with a count."""
+
+    """
+    This table has just now been read from the Glue table catalog
+    and has not yet had any transformations done to it.
+    """
+    READ = "READ"
```
BryanFauble, line 423: The docstrings need to go below the thing it's referencing: https://peps.python.org/pep-0224/

If it does not than the hover document is incorrect:
image

philerooski: Oh, that's an interesting choice.

src/glue/jobs/s3_to_json.py
philerooski: Not in the standard library afaik. IMO, it's not worth adding another dependency.

PR #95, opened 2023-12-14
[ETL-585] Upgrade worker type for HealthKitV2Samples and FitbitIntradayCombined

philerooski:


thomasyu888 (approved):
🔥 This looks good, but should the "WorkerType" just be configurable and be something you can pass through a configuration instead of this?

philerooski:
> 🔥 This looks good, but should the "WorkerType" just be configurable and be something you can pass through a configuration instead of this?

@thomasyu888 We use Jinja in the template to dynamically generate each of these jobs based on the data types we have available in `src/glue/resources/table_columns.yaml`. So the configuration file specifies the properties shared in common by all jobs, but any type-specific logic needs to go in the template.

PR #102, opened 2024-01-10
[ETL-590] Fitbit ecg

BryanFauble:
**Problem:**
We are not ingesting fitbit ecg data in the recover pipeline

**Solution:**
Adding the new table schema and index field

**Testing:**
See the testing completed in the jira comments https://sagebionetworks.jira.com/browse/ETL-590 - I tested this against a sample data file provided to us and verified that we receive a parquet data file with the expected fields after running the pipeline.


rxu17 (approved):
LGTM!

PR #104, opened 2024-02-08
[ETL-610] Add SQS queue for input to raw buckets processing

philerooski:
Most of the changes involve renaming some parameters and stack names related to our original SQS queue (the one that's a go between for the input bucket and S3 to JSON Glue workflow). This was originally just named `sqs-queue` (stack name `sqs-S3ToLambda`), so I renamed the stack name and the file itself to `sqs-input-to-intermediate` to differentiate it from the new `sqs-input-to-raw` stack.


rxu17 (approved):
LGTM!

PR #106, opened 2024-02-21
[ETL-621] Add SNS topic to broadcast new exports

philerooski:
Lot's of files touched here, but essentially I've inserted an SNS topic between our S3 input bucket and our SQS queues (now there's two queues!) in anticipation of triggering both input -> intermediate and input -> raw processes in reaction to new export data. That looks sort of like this:

image

* `templates/sqs-queue.yaml` was modified to subscribe to the SNS topic
* `src/lambda_function/s3_to_glue/app.py` was modified to work with events formatted as `SQS(SNS(S3))`
* `config/prod/namespaced/s3-event-config-lambda.yaml` was modified to add the S3 event notification for the SNS topic, rather than the SQS queue.

Renamed files:
* `config/develop/namespaced/s3-event-config-lambda.yaml` renamed to `config/develop/namespaced/lambda-s3-event-config.yaml` to match the pattern `{service_name}-{descriptor}`
* `config/develop/namespaced/s3-event-config-lambda-role.yaml` renamed to `config/develop/namespaced/lambda-s3-event-config-role.yaml`
* etc. for `prod`

New files:
* `config/develop/namespaced/sns-topic.yaml`
* `config/prod/namespaced/sns-topic.yaml`
* `templates/sns-topic.yaml`


BryanFauble (approved):
🔥 Changes look good to me!

rxu17 (approved):
LGTM! Were some of the main reasons for us moving from S3 -> SQS to S3 -> SNS -> SQS due to being able to decouple our S3 bucket from SQS (given we have multiple now) and make it easier to manage (we don't have to modify S3 bucket configuration directly as much if anything changes on the SQS queue side).

philerooski:
@rxu17 
> Were some of the main reasons for us moving from S3 -> SQS to S3 -> SNS -> SQS due to being able to decouple our S3 bucket from SQS (given we have multiple now) and make it easier to manage

No, it's not possible to have S3 send event notifications to more than one destination for the same S3 prefix. I'm not sure why this is, but maybe they don't want to have to support SNS-like capabilities for every service they implement event hooks for.

rxu17:
> No, it's not possible to have S3 send event notifications to more than one destination for the same S3 prefix. I'm not sure why this is, but maybe they don't want to have to support SNS-like capabilities for every service they implement event hooks for.

I thought we could configure multiple notification configs to the S3 bucket with the same s3 prefix. Bummer.

src/lambda_function/s3_to_glue/events/generate_test_event.py
```diff
@@ -183,9 +190,33 @@ def create_sqs_message(s3_event_record: dict) -> dict:
       "eventSourceARN": "arn:aws:sqs:us-east-1:914833433684:mynamespace-sqs-S3ToLambda-Queue",
       "awsRegion": "us-east-1"
     }
-    sqs_event_record["body"] = json.dumps({"Records": [s3_event_record]})
+    sqs_event_record["body"] = json.dumps(sns_notification)
     return sqs_event_record
 
+def create_sns_notification(s3_event_record):
```
thomasyu888, line 196: Nit: typing hinting

thomasyu888 (approved):
🔥 LGTM! just a tiny Nit, but going to approve.

PR #108, opened 2024-03-01
[ETL-628] Rework S3 event config lambda

philerooski:
Significant changes in the S3 event config lambda to support indexing individual configurations based on their S3 prefix/suffix overlap.

`.github/workflows/upload-and-deploy-to-prod-main.yaml` and `.github/workflows/upload-and-deploy.yaml`
* Delete the potentially preexisting notification configuration for a namespace before creation.

`src/lambda_function/s3_event_config/app.py`
* We introduce three new classes that abstract the API objects relevant to S3 event notifications:
  - `NotificationConfiguration`: creates a common interface for interacting with individual notification configurations.
  - `NotificationConfigurationType`: subsets the Enum class to restrict the types allowed for a `NotificationConfiguration` to `Queue`, `Topic`, and `LambdaFunction`.
  - `BucketNotificationConfigurations`: A convenience class which wraps a collection of `NotificationConfiguration` objects and provides a method for converting them back into a dict representation which can be used with `put_bucket_notification_configuration`.
* We change the behavior of some functions:
  - `add_notification` is now more strict. Notification configurations can be added, but not updated. See the docstring for more details.
  - `delete_notification` now uses the prefix filter rules to identify which notification needs deletion.

### DIFF
--- a/.github/workflows/upload-and-deploy.yaml
+++ b/.github/workflows/upload-and-deploy.yaml
@@ -233,7 +233,18 @@ jobs:
       - name: "Deploy sceptre stacks to dev"
         run: pipenv run sceptre --var "namespace=${{ env.NAMESPACE }}" launch develop --yes
 
-      - name: Configure S3 to Glue lambda with S3 trigger
+      - name: Delete preexisting S3 event notification for this namespace
+        uses: gagoar/invoke-aws-lambda@v3
+        with:
+          AWS_ACCESS_KEY_ID: ${{ env.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ env.AWS_SECRET_ACCESS_KEY }}
+          AWS_SESSION_TOKEN: ${{ env.AWS_SESSION_TOKEN }}
+          REGION: ${{ env.AWS_REGION }}
+          FunctionName: ${{ env.NAMESPACE }}-S3EventConfig
+          Payload: '{"RequestType": "Delete"}'
+          LogType: Tail
+
+      - name: Create S3 event notification for this namespace
         uses: gagoar/invoke-aws-lambda@v3
         with:
           AWS_ACCESS_KEY_ID: ${{ env.AWS_ACCESS_KEY_ID }}
--- a/src/lambda_function/s3_event_config/app.py
+++ b/src/lambda_function/s3_event_config/app.py
@@ -71,11 +71,11 @@ class NotificationConfiguration:
         """
         Assign the ARN of this notification configuration to the `arn` property.
         """
-        if self.type == "Topic":
+        if self.type == NotificationConfigurationType.Topic.value:
             arn = self.value["TopicArn"]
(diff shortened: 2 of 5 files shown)


rxu17 (approved):
LGTM!

src/lambda_function/s3_event_config/app.py
```diff
@@ -48,14 +49,61 @@ def lambda_handler(event, context):
...
+    def __init__(self, notification_type: NotificationConfigurationType, value: dict):
+        self.type = notification_type.value
+        self.value = value
+        self.arn = self.get_arn()
+
+    def get_arn(self):
+        """
+        Assign the ARN of this notification configuration to the `arn` property.
+        """
+        if self.type == "Topic":
```
BryanFauble, line 74: nit: Since you have an enum you could do:
`if self.type == NotificationConfigurationType.Topic.value`

To prevent repeating the string

src/lambda_function/s3_event_config/app.py
```diff
@@ -66,77 +114,71 @@ def get_existing_bucket_notification_configuration_and_type(
...
 
 
-def get_matching_notification_configuration(
-    destination_type_arn: str,
-    existing_notification_configurations_for_type: list,
-    destination_arn: str,
-) -> MatchingNotificationConfiguration:
+def get_notification_configuration(
+    bucket_notification_configurations: BucketNotificationConfigurations,
+    bucket_key_prefix: typing.Union[str,None]=None,
```
BryanFauble, line 140: nit: Consistency for whitespace:
```
typing.Union[str,None]
vs
typing.Union[NotificationConfiguration, None]
```

.github/workflows/upload-and-deploy.yaml
```diff
@@ -233,7 +233,18 @@ jobs:
       - name: "Deploy sceptre stacks to dev"
         run: pipenv run sceptre --var "namespace=${{ env.NAMESPACE }}" launch develop --yes
 
-      - name: Configure S3 to Glue lambda with S3 trigger
+      - name: Delete preexisting S3 event notification for this namespace
```
BryanFauble, line 236: Does anything need to be done in the cleanup workflow? https://github.com/Sage-Bionetworks/recover/blob/058b0779c7692055eecd1e6bd1e6fd3bfe7aa704/.github/workflows/cleanup.yaml#L4

BryanFauble (approved):
Just a few small comments! Otherwise LGTM!

src/lambda_function/s3_event_config/app.py
philerooski: Initially, this seemed like a more complicated way of achieving the same result. But I like knowing that an IDE would raise a warning on this line if we ever removed support for one of these types.

.github/workflows/upload-and-deploy.yaml
philerooski: Good idea. We will delete the namespace's event configuration from the input bucket.

PR #110, opened 2024-04-10
[ETL-631] Update index fields with `ParticipantIdentifier` and propagate participant id fields

philerooski:
* Update index fields
* If parent parquet dataset has `ParticipantID` field, propagate that to the child
* Small change to the FitbitEcg schema to include the export start and end date fields

### DIFF
--- a/src/glue/jobs/json_to_parquet.py
+++ b/src/glue/jobs/json_to_parquet.py
@@ -37,42 +37,42 @@ INDEX_FIELD_MAP = {
     "enrolledparticipants": ["ParticipantIdentifier"],
     "fitbitprofiles": ["ParticipantIdentifier", "ModifiedDate"],
     "fitbitdevices": ["ParticipantIdentifier", "Date"],
-    "fitbitactivitylogs": ["LogId"],
+    "fitbitactivitylogs": ["ParticipantIdentifier", "LogId"],
     "fitbitdailydata": ["ParticipantIdentifier", "Date"],
-    "fitbitecg": ["FitbitEcgKey"],
+    "fitbitecg": ["ParticipantIdentifier", "FitbitEcgKey"],
     "fitbitintradaycombined": ["ParticipantIdentifier", "Type", "DateTime"],
     "fitbitrestingheartrates": ["ParticipantIdentifier", "Date"],
-    "fitbitsleeplogs": ["LogId"],
-    "healthkitv2characteristics": ["HealthKitCharacteristicKey"],
-    "healthkitv2samples": ["HealthKitSampleKey"],
-    "healthkitv2heartbeat": ["HealthKitHeartbeatSampleKey"],
-    "healthkitv2statistics": ["HealthKitStatisticKey"],
-    "healthkitv2clinicalrecords": ["HealthKitClinicalRecordKey"],
-    "healthkitv2electrocardiogram": ["HealthKitECGSampleKey"],
-    "healthkitv2workouts": ["HealthKitWorkoutKey"],
-    "healthkitv2activitysummaries": ["HealthKitActivitySummaryKey"],
-    "garminactivitydetailssummary": ["ParticipantID", "SummaryId"],
-    "garminactivitysummary": ["ParticipantID", "SummaryId"],
-    "garminbloodpressuresummary": ["ParticipantID", "SummaryId"],
-    "garmindailysummary": ["ParticipantID", "StartTimeInSeconds"],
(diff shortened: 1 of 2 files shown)


src/glue/jobs/json_to_parquet.py
```diff
@@ -37,42 +37,42 @@
...
-    "garminactivitydetailssummary": ["ParticipantID", "SummaryId"],
-    "garminactivitysummary": ["ParticipantID", "SummaryId"],
-    "garminbloodpressuresummary": ["ParticipantID", "SummaryId"],
-    "garmindailysummary": ["ParticipantID", "StartTimeInSeconds"],
-    "garminepochsummary": ["ParticipantID", "SummaryId"],
-    "garminhealthsnapshotsummary": ["ParticipantID", "StartTimeInSeconds"],
-    "garminhrvsummary": ["ParticipantID", "StartTimeInSeconds"],
-    "garminmanuallyupdatedactivitysummary": ["ParticipantID", "SummaryId"],
-    "garminmoveiqactivitysummary": ["ParticipantID", "SummaryId"],
-    "garminpulseoxsummary": ["ParticipantID", "SummaryId"],
```
thomasyu888, line 63: Was `ParticipantID` the wrong key?

thomasyu888 (approved):
🔥 Thanks for the quick work here!

src/glue/jobs/json_to_parquet.py
```diff
@@ -582,12 +585,11 @@ def add_index_to_table(
...
         logger.info(f"Adding index to {original_field_name}")
-        parent_index = parent_table.select(
-            (
-                [selectable_original_field_name, "cohort"]
-                + INDEX_FIELD_MAP[table_data_type]
-            )
-        ).distinct()
+        index_fields = INDEX_FIELD_MAP[table_data_type]
+        additional_fields = [selectable_original_field_name, "cohort"]
+        if "ParticipantID" in parent_table.columns:
```
rxu17, line 590: Should we modify the test for this function to check this part that ParticipantID gets included in `additional_fields` if it exists?

src/glue/jobs/json_to_parquet.py
```diff
@@ -582,12 +585,11 @@ def add_index_to_table(
         else:
             selectable_original_field_name = original_field_name
         logger.info(f"Adding index to {original_field_name}")
-        parent_index = parent_table.select(
-            (
-                [selectable_original_field_name, "cohort"]
-                + INDEX_FIELD_MAP[table_data_type]
-            )
-        ).distinct()
+        index_fields = INDEX_FIELD_MAP[table_data_type]
```
rxu17, line 588: Given that now `Participant_Identifier` is a required index field for every data type, I'm thinking maybe we could have a test to ensure that `Participant_Identifier` is in all the keys in the dict of `INDEX_FIELD_MAP`. This is more of future proofing/double checking that we don't modify code that would accidentally affect this. Thoughts?

rxu17 (approved):
LGTM! Thanks for looking into this! Just a few comments.

src/glue/jobs/json_to_parquet.py
philerooski: There's a one-to-one mapping from `ParticipantIdentifier` to `ParticipantID`. They are two different ways of representing the same identifier.

`ParticipantIdentifier` is the only field present in every data type (including SymptomLog) and the CE docs suggest it's the more "official" identifier -- kind of like `HealthCode` to `ExternalId` in mPower.

src/glue/jobs/json_to_parquet.py
philerooski: Ahh, yes we should. I'm always forgetting about tests 🤦

philerooski: > why would both ParticipantID and ParticipantIdentifier exist in a dataset?

I don't know the exact reason, but it's not unusual to have one be the "global" identifier and the other to be a study or app specific identifier.

src/glue/jobs/json_to_parquet.py
philerooski: I don't want to assume that `ParticipantIdentifier` will be included with every data type going forward. That's a CE decision, and why I explicitly included the field in the `INDEX_FIELD_MAP` rather than specifying it once in this function.

src/glue/jobs/json_to_parquet.py
rxu17: That's not confusing at all :D

PR #112, opened 2024-04-26
[IBCDPE-923] snowflake integration

thomasyu888:
**Problem**
Currently, it is difficult to query the RECOVER dataset.  It would be much easier to do so with snowflake in the mix.

**Solution**
Follow instructions here: https://docs.snowflake.com/en/user-guide/data-load-s3-config-storage-integration to integrate RECOVER S3 buckets with snowflake

Had the close the old PR because it contained capitol letters in the branch name.


BryanFauble (approved):
LGTM!

PR #113, opened 2024-05-08
[ETL-644] Update `drop_table_duplicates` function to use new method

philerooski:
Primary changes are L192-208 where we swap the old (and faulty) sort + drop duplicate method of removing duplicates with the new window + sort + rank + filter method.

I also did some refactoring to make things a little more intuitive:
* `drop_table_duplicates` and `drop_deleted_healthkit_data` now return Spark DataFrames rather than Glue DynamicFrames. These functions already produced DataFrames so I removed the final cast as a DynamicFrame to simplify things. We cast back to a DynamicFrame on L662 since both branches of the conditional (relationalize+write else write) take a DynamicFrame as input.
* Mostly as a consequence of the above refactor, rather than deriving the data type from the table name in non-`main` functions I derive the data type once in `main` and pass that as an argument to functions which reference the data type. I think this consolidates the number of places where a data type is derived from its table name (or from the job arguments, where the table name is derived), making things more consistent generally.

### DIFF
--- a/src/glue/jobs/json_to_parquet.py
+++ b/src/glue/jobs/json_to_parquet.py
@@ -24,6 +24,9 @@ from awsglue.job import Job
 from awsglue.gluetypes import StructType
 from awsglue.utils import getResolvedOptions
 from pyspark import SparkContext
+from pyspark.sql import Window
+from pyspark.sql.functions import row_number, col
+from pyspark.sql.dataframe import DataFrame
 
 # Configure logger to use ECS formatting
 logger = logging.getLogger(__name__)
@@ -122,7 +125,7 @@ def get_table(
     glue_context: GlueContext,
     record_counts: dict,
     logger_context: dict,
-) -> DynamicFrame:
+    ) -> DynamicFrame:
     """
     Return a table as a DynamicFrame with an unambiguous schema. Additionally,
     we drop any superfluous partition_* fields which are added by Glue.
@@ -162,11 +165,10 @@ def get_table(
 
 def drop_table_duplicates(
     table: DynamicFrame,
-    table_name: str,
-    glue_context: GlueContext,
+    data_type: str,
     record_counts: dict[str, list],
     logger_context: dict,
-) -> DynamicFrame:
+    ) -> DataFrame:
     """
     Drop duplicate samples and superflous partition columns.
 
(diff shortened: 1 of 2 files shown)


BryanFauble (approved):
Excellent work!

src/glue/jobs/json_to_parquet.py
```diff
@@ -177,49 +179,52 @@ def drop_table_duplicates(
...
-    table = DynamicFrame.fromDF(
-        dataframe=(
-            sorted_spark_df.dropDuplicates(subset=INDEX_FIELD_MAP[table_data_type])
-        ),
-        glue_ctx=glue_context,
-        name=table_name,
+        window_ordered = window_unordered.orderBy(
+                col("export_end_date").desc()
+        )
+    table_no_duplicates = (
```
thomasyu888, line 203: Fun fact, this is exactly what I do to create de-duplicated tables within the synapse data warehouse: https://github.com/Sage-Bionetworks/snowflake/blob/b4755ad226abe29d30308708539c99d632dd2337/synapse_data_warehouse/synapse/tables/V1.11.0__initial_creation_of_latest_tables.sql#L3-L17

thomasyu888 (approved):
🔥 LGTM!

src/glue/jobs/json_to_parquet.py
philerooski: I think this way may be more common in databases, versus sorting and dropping as you would an in-memory dataframe. Perhaps it has something to do with the distributed (across partitions) nature of the data.

philerooski:
As a sanity check, I compared the Fitbit Intraday Combined Parquet data produced by the main branch (which uses the older method of dropping duplicates) to this feature branch. Both were produced from the same set of JSON pilot data in the dev environment. While each contained the same number of unique records, I saw differences in which exports the records were sourced from, confirming that these two methods were dropping duplicates differently:
```
>>> fitbit_branch.export_end_date.value_counts() 
export_end_date
2023-01-12T00:00:00    1996228
2023-01-03T00:00:00     173251
2023-01-14T00:00:00      59005
2023-06-27T00:00:00      10050
Name: count, dtype: int64
>>> fitbit_main.export_end_date.value_counts()
export_end_date
2023-01-12T00:00:00    1918724
2023-01-03T00:00:00     251167
2023-01-14T00:00:00      58593
2023-06-27T00:00:00      10050
Name: count, dtype: int64
``` 
I then read the JSON data as a pandas dataframe and dropped duplicates. I understand and trust how duplicates are dropped in an in-memory dataframe more so than in a Spark dataframe, so this just a way to independently verify that we are dropping duplicates correctly. I saw the same distribution of records as the feature branch:
```
>>> fitbit_json_df_no_dups.export_end_date.value_counts()
2023-01-12T00:00:00    1996228
2023-01-03T00:00:00     173251
2023-01-14T00:00:00      59005
2023-06-27T00:00:00      10050
Name: export_end_date, dtype: int64
```

PR #115, opened 2024-05-30
[ETL-647] Only submit most recent ~2 weeks of exports for integration tests

philerooski:
* Rather than copying over all input data, only copy the 28 (~2 weeks from each cohort) most recent exports to the namespaced location in the input bucket.
* The S3 URI of these 28 exports is uploaded to the S3 cloudformation bucket under the key `{namespace}/integration_test_exports.json` as a JSON array. This is for later reference in the comparison jobs. e.g.,
```
[
  "s3://recover-dev-input-data/pilot-data/adults_v1/2023-01-12T22--02--17Z_77fefff8-b0e2-4c1b-b0c5-405554c92368",
  "s3://recover-dev-input-data/pilot-data/adults_v1/2023-01-03T20--19--27Z_6fb88c91-eb34-4278-a603-4de03dfd9c70",
  "s3://recover-dev-input-data/pilot-data/pediatric_v1/2023-06-26T22_35_22Z_8ae9d391-5c5d-4d79-bdce-6d86314caf8c",
 ...
]
```
* We now initiate integration tests for staging data by copying over exports from the main namespace, rather than creating and submitting test events.

### DIFF
--- a/.github/workflows/upload-and-deploy.yaml
+++ b/.github/workflows/upload-and-deploy.yaml
@@ -22,7 +22,6 @@ jobs:
       - uses: actions/setup-python@v4
       - uses: pre-commit/action@v3.0.0
 
-
   upload-files:
     name: Upload files to S3 bucket in development
     runs-on: ubuntu-latest
@@ -265,6 +264,8 @@ jobs:
     permissions:
       id-token: write
       contents: read
+    env:
+      EXPORT_S3_KEY_PREFIX: pilot-data
     steps:
       - name: Setup code, pipenv, aws
         uses: Sage-Bionetworks/action-pipenv-aws-setup@v3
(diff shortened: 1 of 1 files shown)


thomasyu888 (approved):
🔥 LGTM!  It does look like there's some duplicated code logic, we could look to make these composite Github actions in the future but that's low priority since thats just aesthetics

.github/workflows/upload-and-deploy.yaml
```diff
@@ -277,11 +278,30 @@ jobs:
         if: github.ref_name != 'main'
         run: echo "NAMESPACE=$GITHUB_REF_NAME" >> $GITHUB_ENV
 
-      - name: Copies over test files from ingestion bucket
+      - name: Fetch the most recent exports.
```
rxu17, line 281: Do we have ~2 weeks of exports from the develop input bucket to fetch? Or will this just get whatever is available even if it's not 2 weeks worth?

rxu17 (approved):
LGTM! Just a couple of Qs

.github/workflows/upload-and-deploy.yaml
```diff
@@ -333,22 +354,27 @@ jobs:
...
+        id: recent-exports
+        run: |
+          # Retrieve the last ~2 weeks of exports from each cohort
+          # Ignore keys which end with "/" and which match "owner.txt"
+          echo "KEYS=$(
+            aws s3api list-objects-v2 \
+            --bucket $PROD_INPUT_BUCKET \
+            --prefix "$EXPORT_S3_KEY_PREFIX/" \
+            --query '((sort_by(Contents[? !ends_with(Key, `/`) && !contains(Key, `owner.txt`)], &LastModified)[::-1])[:28])[*].Key' |
```
rxu17, line 366: If we happen to have more cohorts in the future or less, what do would modify here to support that? 

I know some data types (like google fit samples?) only has adult data, so then based on this logic would this pull the 28 days of adult data?

.github/workflows/upload-and-deploy.yaml
philerooski: Yeah, it acts like a list slice in Python `[:28]`.

rxu17: Nit: Gotcha, should we set this as an env var? so this way we can use it in both the two integration test for staging and development, and easier to change in the future.

.github/workflows/upload-and-deploy.yaml
philerooski: >If we happen to have more cohorts in the future or less, what do would modify here to support that?

Assuming that these new cohorts are also exported on a daily basis, we would still see equal representation of the cohorts in the 28 most recent exports. 28 is more or less an arbitrary amount to submit for integration tests, so having fewer `pediatric_v1` exports, for example, is not a big deal. But we could increase this number proportionally to the number of new cohorts if we like.

I don't think the number of cohorts would decrease, since we want to support the processing of data from any cohort in perpetuity.

> I know some data types (like google fit samples?) only has adult data, so then based on this logic would this pull the 28 days of adult data?

I'm not sure what you mean. This code pulls the 28 most recent ZIP exports across cohorts, so if there's a data type that's only present in one of the cohorts than we would expect to see fewer JSON files of that data type.

.github/workflows/upload-and-deploy.yaml
philerooski: Ohh that's a good idea. I accidentally merged before reading this comment. I can fix as part of #117

PR #118, opened 2024-06-04
[ETL-656] Limit input-to-intermediate Lambda batch size to 1

philerooski:
See ticket for rationale.

### DIFF
--- a/src/lambda_function/s3_to_glue/template.yaml
+++ b/src/lambda_function/s3_to_glue/template.yaml
@@ -27,13 +27,13 @@ Parameters:
 
   LambdaBatchSize:
     Type: Number
-    Default: 10
+    Default: 1
     Description: >-
       The maximum number of S3 messages in a SQS event that Lambda will process in a batch
 
   LambdaMaximumBatchingWindowInSeconds:
     Type: Number
-    Default: 300
+    Default: 60
     Description: >-
       The maximum amount of time in seconds Lambda will batch messages before polling
       the SQS queue and processing them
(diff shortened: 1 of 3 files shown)


src/lambda_function/s3_to_glue/template.yaml
rxu17, line 1: You mentioned the bottleneck in our integration testing pipeline: If too many S3 events (new exports) are consumed by the S3 to JSON job, it will timeout. Are you saying that 10 events in a bundle will be too much for the S3 to JSON job?

src/lambda_function/s3_to_glue/template.yaml
```diff
@@ -27,13 +27,13 @@ Parameters:
...
     Type: Number
-    Default: 10
+    Default: 1
     Description: >-
       The maximum number of S3 messages in a SQS event that Lambda will process in a batch
 
   LambdaMaximumBatchingWindowInSeconds:
     Type: Number
-    Default: 300
+    Default: 60
```
rxu17, line 36: Yeah it seems like we really won't need this anymore since our S3 to JSON job has a maximum concurrency of ~150 to begin with.

rxu17 (approved):
LGTM!

src/lambda_function/s3_to_glue/template.yaml
philerooski: Yes, depending on which exports it's trying to process. The timeout is 2 hours, so more than 6 adult exports would be likely to cause a timeout.

I could have simply increased the timeout, but limiting the batch size to 1 will allow us to process exports in parallel and complete our tests faster. Most of the cost-efficiency gains from a larger batch size come from the reduced setup time of worker instances -- which is a pretty negligible component of the total processing time if it takes us 20 minutes to process one export.

src/lambda_function/s3_to_glue/template.yaml
philerooski: Nope, it doesn't make much sense to set it at all. But it doesn't hurt to keep it explicitly defined in the template, either.

PR #119, opened 2024-06-14
[ETL-648] Modify comparsion job to only consider records from recent exports

rxu17:
# Purpose
The purpose of this PR is to update the comparison job that compares parquet datasets between `staging` and `main` so that it incorporates the changes in #115 where we only submit the most recent ~2 weeks of exports for integration testing of running our staging pipeline all the way through. 

See [JIRA Ticket](https://sagebionetworks.jira.com/browse/ETL-648) for the full description and reasoning but essentially we pull the data type / cohort / export_end_date from the JSON array that contains the exports to be used in the integration test: `s3://{cloudformation_bucket_name}/{namespace}/integration_test_exports.json`, then use those values to filter the production parquet datasets (the ones in `main` namespace) before running the comparisons.

There's also an extra layer of difficulty because we don't currently support comparisons of subtype datasets just the top level datasets (and now deleted datasets)

# Changes
There's been a lot of maintenance changes other than the actual feature changes.

### Helper Functions
Addition of helper functions to help parse the exports for the filter values we need in order to filter our `main` dataset:
- [get_export_end_date](https://github.com/Sage-Bionetworks/recover/blob/etl-648/src/glue/jobs/compare_parquet_datasets.py#L230)
- [get_cohort_from_s3_uri](https://github.com/Sage-Bionetworks/recover/blob/etl-648/src/glue/jobs/compare_parquet_datasets.py#L263)
- [get_data_type_from_filename](https://github.com/Sage-Bionetworks/recover/blob/etl-648/src/glue/jobs/compare_parquet_datasets.py#L284)

A bunch of the code comes from [s3_to_json.py](https://github.com/Sage-Bionetworks/recover/blob/etl-648/src/glue/jobs/s3_to_json.py) but I only pulled out the snippets we need for the scope of this feature.

### Main feature functions:

1. [get_integration_test_exports_json](https://github.com/Sage-Bionetworks/recover/blob/etl-648/src/glue/jobs/compare_parquet_datasets.py#L349)
2. [get_json_files_in_zip_from_s3](https://github.com/Sage-Bionetworks/recover/blob/etl-648/src/glue/jobs/compare_parquet_datasets.py#L316)
3. [get_exports_filter_values](https://github.com/Sage-Bionetworks/recover/blob/etl-648/src/glue/jobs/compare_parquet_datasets.py#L373)
4. [convert_filter_values_to_expression](https://github.com/Sage-Bionetworks/recover/blob/etl-648/src/glue/jobs/compare_parquet_datasets.py#L419)
5. [get_parquet_dataset](https://github.com/Sage-Bionetworks/recover/blob/etl-648/src/glue/jobs/compare_parquet_datasets.py#L455)
This was a **pre-existing** function and has been redesigned. Note that we use `pyarrow`'s dataset feature for reading in the `main` dataset because we want to utilize the predicate pushdown feature of `pyarrow.datasets.dataset` which allows us to filter data early in the reading process, which can significantly improve performance by reducing I/O. This is helpful as our comparison jobs are limited in memory available given it's just a Python glue job.

###  From previous PRs
- Need to update the code to account for the update of the index fields with `ParticipantIdentifier` in #110 
- Need to update the code to account for the update of added deleted tables to our table schema in #116

We needed to update the `INDEX_FIELD_MAP` in this PR to reflect those changes and also update all our staging dataset vs main dataset test examples to incorporate those changes (e.g: `ParticipantIdentifier` is a required index column now for all data types thus it is used to join the staging dataset and main datasets so they are comparable)

### Mocking pyarrow's fileystem object using moto server
- Need to have a way to properly mock and test the [pyarrow S3 Filesystem object] (https://arrow.apache.org/docs/python/generated/pyarrow.fs.S3FileSystem.html) (this wasn't available before)

We also added two new libraries: `flask` and `flask-cors` to the extra libraries for the compare parquet glue job and python shell glue docker as they are required dependencies of using the [moto server](https://docs.getmoto.org/en/latest/docs/server_mode.html). This server is the only way to test using a filesystem object to read parquet from mocked s3 buckets (ref: https://github.com/apache/arrow/issues/31811). 

The function that use this moto server test are (and its respective unit tests): 
- [get_parquet_dataset](https://github.com/Sage-Bionetworks/recover/blob/etl-648/src/glue/jobs/compare_parquet_datasets.py#L455) and its [unit tests](https://github.com/Sage-Bionetworks/recover/blob/etl-648/tests/test_compare_parquet_datasets.py#L456-L668)

###  Redesigning how we check if a namespace has data of a specific data type
Thanks to [discussion here](https://github.com/Sage-Bionetworks/recover/pull/119#discussion_r1648030195), we decided to re-do the check for a data type in a specific namespace. We have removed all related functions as mentioned in the discussion and replaced it with the new function:
- [has_parquet_files](https://github.com/Sage-Bionetworks/recover/blob/etl-648/src/glue/jobs/compare_parquet_datasets.py#L745)
This is used in main to check that both the "staging" and "main" namespaces have parquet data before running the comparison reports

Prior to merging, we will need to clear the staging namespace in the `recover-intermediate-data` bucket

###  Refactoring `main` function
We had a lot of duplicated logic in the `main` function prior ([see comment here](https://github.com/Sage-Bionetworks/recover/pull/119#pullrequestreview-2131182682)) with the `s3.put_object` calls for every kind of report we wanted to upload to s3. That logic has been refactored in `main` alongside a new helper function to upload to s3:
- [upload_reports_to_s3](https://github.com/Sage-Bionetworks/recover/blob/etl-648/src/glue/jobs/compare_parquet_datasets.py#L774)

# Testing

- [x] Unit tests
- [x] Integration testing on `dev` pipeline
This runs successfully on the pipeline on the workflow: `wr_77376de723cc0781cefc0d2775a2e19cc269f54b020ae4e774637eb32b144050`. The comparison txt files are outputted to `s3://recover-dev-processed-data/etl-648/comparison_result/` and the export_end_dates for the main vs staging datasets match up. The comparison files show the expected differences between `etl-648` namespace and `main` namespace. See [JIRA Ticket comment](https://sagebionetworks.jira.com/browse/ETL-648?focusedCommentId=211429) for more details on this.

### DIFF
--- a/src/glue/jobs/compare_parquet_datasets.py
+++ b/src/glue/jobs/compare_parquet_datasets.py
@@ -18,8 +18,8 @@ from pyarrow import fs
 logger = logging.getLogger()
 logger.setLevel(logging.INFO)
 
-COHORTS = {"adults": "adults_v1", "pediatric": "pediatric_v1"}
-
+ADULTS = "adults_v1"
+PEDIATRIC = "pediatric_v1"
 INDEX_FIELD_MAP = {
     "dataset_enrolledparticipants": ["ParticipantIdentifier"],
     "dataset_fitbitprofiles": ["ParticipantIdentifier", "ModifiedDate"],
@@ -285,10 +285,10 @@ def get_cohort_from_s3_uri(s3_uri: str) -> Union[str, None]:
         Union[str, None]: the cohort if it exists
     """
     cohort = None
-    if COHORTS["adults"] in s3_uri:
-        cohort = COHORTS["adults"]
-    elif COHORTS["pediatric"] in s3_uri:
-        cohort = COHORTS["pediatric"]
+    if ADULTS  in s3_uri:
+        cohort = ADULTS
+    elif PEDIATRIC  in s3_uri:
+        cohort = PEDIATRIC
     else:
         pass
     return cohort
--- a/src/glue/jobs/compare_parquet_datasets.py
+++ b/src/glue/jobs/compare_parquet_datasets.py
@@ -418,7 +418,7 @@ def get_exports_filter_values(
 
 
 def convert_filter_values_to_expression(filter_values: Dict[str, str]) -> ds.Expression:
-    """ Converts the dict of the keys, values to filter on
+    """Converts the dict of the keys, values to filter on
         into filter conditions in the form of a pyarrow.dataset.Expression object
 
         The expression object takes the following structure for a single condition:
@@ -747,28 +747,24 @@ def upload_reports_to_s3(
     s3: boto3.client,
     reports: List[NamedTuple],
     data_type: str,
-    parquet_bucket_name: str,
+    parquet_bucket: str,
     staging_namespace: str,
 ) -> None:
-    """Uploads the various comparison reports to S3 bucket. We currently
-        support the following reports:
-            - main datacompy comparison report
-            - row differences in staging dataset
-            - row differences in main dataset
-            - duplicated data in staging dataset
-            - duplicated data in main dataset
+    """Uploads the various comparison reports to S3 bucket.
 
     Args:
-        s3 (boto3.client): _description_
-        compare_dict (Dict[str, datacompy.Compare]): _description_
-        data_type (str): _description_
-        parquet_bucket_name (str): _description_
-        staging_namespace (str): _description_
+        s3 (boto3.client): s3 client connection
+        reports (List[NamedTuple]): List of report which contain content(str) and
+        file_name(str). Content is the string body of the report and file_name is the
+        name of the file to be saved to S3.
+        data_type (str): data type to be compared for the given datasets
+        parquet_bucket (str): name of the bucket containing the parquet datasets
+        staging_namespace (str): name of namespace containing the "new" data
     """
     for report in reports:
         if report.content:
             s3.put_object(
-                Bucket=parquet_bucket_name,
+                Bucket=parquet_bucket,
                 Key=get_s3_file_key_for_comparison_results(
                     staging_namespace=staging_namespace,
                     data_type=data_type,
--- a/src/glue/jobs/compare_parquet_datasets.py
+++ b/src/glue/jobs/compare_parquet_datasets.py
@@ -1,15 +1,20 @@
 import argparse
-from io import StringIO
+import datetime
+from io import BytesIO, StringIO
 import json
 import logging
 import os
 import sys
+from typing import Dict, List, Union
+import zipfile
 
 from awsglue.utils import getResolvedOptions
 import boto3
 import datacompy
 import pandas as pd
 from pyarrow import fs
+import pyarrow.compute as pc
+import pyarrow.dataset as ds
 import pyarrow.parquet as pq
 
 logger = logging.getLogger()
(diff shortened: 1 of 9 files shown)


src/glue/jobs/compare_parquet_datasets.py
```diff
@@ -150,6 +238,229 @@ def get_S3FileSystem_from_session(
...
+    if len(filename_components) <= 1:
+        export_end_date = None
+    else:
+        if "-" in filename_components[-1]:
+            _, end_date = filename_components[-1].split("-")
+            end_date = datetime.datetime.strptime(end_date, "%Y%m%d")
+        else:
+            end_date = datetime.datetime.strptime(filename_components[-1], "%Y%m%d")
+        export_end_date = end_date.isoformat()
+    return export_end_date
```
BryanFauble, line 266:
```suggestion
    if (
        filename is None
        or filename == ""
        or len((filename_components := os.path.splitext(filename)[0].split("_"))) <= 1
    ):
        return None

    if "-" in filename_components[-1]:
        _, end_date = filename_components[-1].split("-")
        end_date = datetime.datetime.strptime(end_date, "%Y%m%d")
    else:
        end_date = datetime.datetime.strptime(filename_components[-1], "%Y%m%d")
    export_end_date = end_date.isoformat()
    return export_end_date
```

BryanFauble: I like to return early in the invalid cases that way the rest of the function doesn't need to be nested within the else.

src/glue/jobs/compare_parquet_datasets.py
```diff
@@ -150,6 +238,229 @@ def get_S3FileSystem_from_session(
...
+    s3://{bucket_name}/{namespace}/{cohort}/{export_name}
+
+    Args:
+        s3_uri (str): the S3 uri
+
+    Returns:
+        Union[str, None]: the cohort if it exists
+    """
+    cohort = None
+    if "adults_v1" in s3_uri:
```
BryanFauble, line 283: nit: Could extract these out to constants at the top of the file

src/glue/jobs/compare_parquet_datasets.py
```diff
@@ -1,25 +1,25 @@
-import argparse
 import datetime
```
BryanFauble, line 1: I recommend using isort to handle import sort order: https://pycqa.github.io/isort/docs/configuration/pre-commit.html

It could be part of the pre-commit to do this for you.

src/glue/jobs/compare_parquet_datasets.py
```diff
@@ -280,10 +285,10 @@ def get_cohort_from_s3_uri(s3_uri: str) -> Union[str, None]:
         Union[str, None]: the cohort if it exists
     """
     cohort = None
-    if "adults_v1" in s3_uri:
-        cohort = "adults_v1"
-    elif "pediatric_v1" in s3_uri:
-        cohort = "pediatric_v1"
+    if COHORTS["adults"] in s3_uri:
```
BryanFauble, line 288: Could simplify this further instead of using a dict to map this:

```

ADULTS = "adults_v1"
PEDIATRIC = "pediatric_v1

    if ADULTS  in s3_uri:
        cohort = ADULTS 
    elif PEDIATRIC  in s3_uri:
        cohort = PEDIATRIC 
```

BryanFauble (approved):
Changes make sense to me! Great work here

src/glue/jobs/compare_parquet_datasets.py
rxu17: Thanks! Handy, just added it

src/glue/jobs/compare_parquet_datasets.py
```diff
@@ -1,47 +1,107 @@
...
 import pyarrow.parquet as pq
+from awsglue.utils import getResolvedOptions
+from pyarrow import fs
 
 logger = logging.getLogger()
 logger.setLevel(logging.INFO)
 
+ADULTS = "adults_v1"
+PEDIATRIC = "pediatric_v1"
 INDEX_FIELD_MAP = {
```
thomasyu888, line 23: Is this only defined here?

src/glue/jobs/compare_parquet_datasets.py
```diff
@@ -150,6 +213,237 @@ def get_S3FileSystem_from_session(
...
+        filename (str): JSON filename
+
+    Returns:
+        Union[str, None]: the data type otherwise returns None as we
+        currently don't support comparingsubtypes in data
+    """
+    filename_components = os.path.splitext(filename)[0].split("_")
+    data_type = filename_components[0]
+
+    if "HealthKitV2" in data_type and filename_components[-2] == "Deleted":
```
thomasyu888, line 294: Are we certain that "Deleted' will always be in that spot of the file?

src/glue/jobs/compare_parquet_datasets.py
```diff
@@ -150,6 +213,237 @@ def get_S3FileSystem_from_session(
...
+    ):
+        return None
+
+    filename_components = os.path.splitext(filename)[0].split("_")
+
+    if "-" in filename_components[-1]:
+        _, end_date = filename_components[-1].split("-")
+        end_date = datetime.datetime.strptime(end_date, "%Y%m%d")
+    else:
+        end_date = datetime.datetime.strptime(filename_components[-1], "%Y%m%d")
```
thomasyu888, line 244: Nit: There's probably some sorcery you can do there to reduce this, BUT i think this is ultimately more readable which is more important.

thomasyu888 (commented):
🔥 Thanks for the great work here! and great reviews!

thomasyu888 (approved):

src/glue/jobs/compare_parquet_datasets.py
rxu17: It is our expected filename format according to the data docs: https://sagebionetworks.jira.com/wiki/spaces/DPE/pages/3062398977/Data+Dictionary#Derivation. But if they ever decide to suddenly change the format, then yes it wouldn't meet the if statement criteria anymore. But this is OK for this code because deleted subtypes aren't in our `INDEX_FIELD_MAP` so it would change to returning `None` which is also okay since we won't be comparing deleted data types.

src/glue/jobs/compare_parquet_datasets.py
rxu17: Yes

philerooski (commented):
Is there a way to split out the series of conditionals on L857-920? Conceptually they feel like they are all doing a similar thing and probably there is a way to cut out 50 lines of code by consolidating the `s3.put_object` calls, all while improving the readability of `main`. I don't think this function needs tests any more than `main` needs tests, but if there's a straightforward way to test that the `put_object` method is called when each of the conditions is true/false that ought to be sufficient. If it's too complicated to mock, don't bother.

src/glue/jobs/compare_parquet_datasets.py
```diff
@@ -150,6 +213,237 @@ def get_S3FileSystem_from_session(
...
+    """
+    filelist = get_integration_test_exports_json(s3, cfn_bucket, staging_namespace)
+    # create the dictionary of export end dates and cohort
+    export_end_date_vals = {}
+    for s3_uri in filelist:
+        json_files = get_json_files_in_zip_from_s3(s3, input_bucket, s3_uri)
+        cur_cohort = get_cohort_from_s3_uri(s3_uri)
+        for json_file in json_files:
+            cur_data_type = get_data_type_from_filename(json_file)
+            cur_export_end_date = get_export_end_date(json_file)
```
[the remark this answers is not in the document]
rxu17: Mainly for clarity of the test cases and docstring as we keep it to doing one function. It's easier to write test cases when a function can be reduced to one purpose.

src/glue/jobs/compare_parquet_datasets.py
```diff
@@ -150,6 +213,237 @@ def get_S3FileSystem_from_session(
...
+            if cur_data_type == data_type:
+                if cur_cohort in export_end_date_vals.keys():
+                    export_end_date_vals[cur_cohort].append(cur_export_end_date)
+                else:
+                    export_end_date_vals[cur_cohort] = [cur_export_end_date]
+            else:
+                continue
+
+    # create the filter using pyarrow and predicate pushdown
```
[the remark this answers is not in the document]
rxu17: Actually yes i forgot this, I could just replace `get_parquet_dataset` with `get_filtered_main_dataset` because if `filter` is None in `get_filtered_main_dataset`, it will just read in the dataset still as a whole with no filter.

Yes originally when I first designed this function, I wrestled between splitting the current `get_exports_filter_values` function into the two you described mainly because I felt like the tests would be sort of redundant between the two because our end goal of the function is that it returns the right ds.Expression (filter and values) (and we already have so many tests). But i can see a use case where we may want to utilize the dict or list[namedtuple] of filter values over a `ds.Expression` object so I'll just split the two functions. Easier to troubleshoot as well in the future.

src/glue/jobs/compare_parquet_datasets.py
```diff
@@ -150,6 +213,237 @@ def get_S3FileSystem_from_session(
...
+        for cohort, values in export_end_date_vals.items():
+            column_filter = (ds.field("cohort") == cohort) & (
+                ds.field("export_end_date").isin(values)
+            )
+            if exports_filter is None:
+                exports_filter = column_filter
+            else:
+                # Combine filters using logical OR
+                exports_filter = exports_filter | column_filter
+    return exports_filter
```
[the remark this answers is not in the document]
rxu17: I will leave it as is for now.

src/glue/jobs/compare_parquet_datasets.py
```diff
@@ -514,6 +820,9 @@ def main():
     if data_types_to_compare:
```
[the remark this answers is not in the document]
rxu17: I believe in the original design, we included this because we wanted to add a line to the current comparison report if `main` or `staging` doesn't have a particular data type parquet folder and the other does (as an indicator or whether they have a data type to compare against). But yes, we probably don't need to include that anymore/it seems like more overkill.

I don't think we would be parsing the file names from the exports. We could use a function that's like `has_files_in_s3` that just parses and checks that there are parquet files in the data type folder for that namespace. And then remove adding that line to the compare parquet report.

Then yes we can remove `get_data_types_to_compare`, `compare_dataset_data_types`, and get_folders_in_s3_bucket`

src/glue/jobs/compare_parquet_datasets.py
```diff
@@ -460,7 +767,7 @@ def compare_datasets_by_data_type(
     # check that they have columns in common to compare
     elif not has_common_cols(staging_dataset, main_dataset):
```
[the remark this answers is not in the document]
rxu17: Right, I forgot why I separated them in the first place (maybe I'll remember when I work on this). I can build them into the same function.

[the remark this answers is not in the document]
rxu17: oh i separated them because `has_common_cols` needs both datasets as input while the `is_valid_dataset` is designed for individual datasets/the namespace they have. I think I'll keep them as is just because that would over complicate that one function (more parameters). I wonder if instead of messages in the comparison report, we should just have them throw exceptions.

src/glue/jobs/compare_parquet_datasets.py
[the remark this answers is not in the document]
rxu17: I opted to just remove the functionality related to getting available data types. It helps clear up a lot of the complexity in `main`

rxu17:
> Is there a way to split out the series of conditionals on L857-920? Conceptually they feel like they are all doing a similar thing and probably there is a way to cut out 50 lines of code by consolidating the s3.put_object calls, all while improving the readability of main. I don't think this function needs tests any more than main needs tests, but if there's a straightforward way to test that the put_object method is called when each of the conditions is true/false that ought to be sufficient. If it's too complicated to mock, don't bother.

Yes, I thought about it and have a potential solution

src/glue/jobs/compare_parquet_datasets.py
philerooski: @thomasyu888 Yes, at least that's the format CE has used so far (they don't provide an API definition or anything). And there is always an underscore in the filename to separate the data type name from the date, so the `.[-2]` element will always exist (same caveat as previous sentence).

philerooski:
> we have to create a fixture that outputs both the mocked with mock_s3 s3 client used to create the bucket and the bucket name we used.

Probably it's a bit cleaner to store the bucket name and any other S3-related references to their own fixtures, e.g.,
```
@pytest.fixture
def s3_bucket_name():
    bucket_name = 'test-bucket'
    yield bucket_name
```
then pass them as input to a `mock_s3_environment` fixture, which sets up the environment and then yields the mocked S3 client.
```
@pytest.fixture
def mock_s3_environment(s3_bucket_name):
    with mock_s3():
        s3 = boto3.client('s3', region_name='us-east-1')
        s3.create_bucket(Bucket=s3_bucket_name)
        yield s3
```
Then you could explicitly pass those references you need (bucket name, for example) to the test function, along with the mocked client, rather than implicitly passing them through a single fixture.
```
def test_function(mock_s3_environment, s3_bucket_name):
    mock_s3_environment.put_object(Bucket=s3_bucket_name, ...)
```

src/glue/jobs/compare_parquet_datasets.py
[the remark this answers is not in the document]
philerooski: re Exceptions: Does failing any one of these checks indicate that there's something seriously wrong with the Parquet data? 
* No columns in one of the datasets (What if the dataset has no rows?)
* Duplicate columns
* No intersecting columns

If so, I think an exception makes sense. But if it's something that could theoretically happen but doesn't indicate any _wrongness_ or other error in the production of the Parquet data then it's better to have the comparison report.

rxu17:
Sounds good. Let me try that out. I thought I originally tried that combo but I must have done it the opposite way (put the bucket creation in the bucket fixture) and that didn't work.

src/glue/jobs/compare_parquet_datasets.py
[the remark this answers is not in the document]
rxu17: > re Exceptions: Does failing any one of these checks indicate that there's something seriously wrong with the Parquet data?

    No columns in one of the datasets (What if the dataset has no rows?)
    Duplicate columns
    No intersecting columns

Yeah I think `get_parquet_dataset` won't throw an exception if the dataset is read in empty or has empty cols (as long as the filter values is `None` as well) Agree those three situations should be exceptions as they shouldn't happen to the data. I'll convert to exceptions.

philerooski (approved):

PR #121, opened 2024-07-02
[ETL-654] Clean up before integration test run

BryanFauble:
**Problem:**

1. When an integration test is run multiple times on a namespaced branch (Or on the staging namespace) it will affect results.

**Solution:**

1. Cleaning out the input bucket, and json intermediate bucket before running the integration test in both namespaced, and main branches

**Testing:**

1. Will be verifying against namespaced branches for testing, and then once merged into the main branch

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -73,8 +73,8 @@ This is responsible for cleaning up any data locations that are used by integrat
 tests during the staging run. This is used after `sceptre-deploy-staging`, but before
 `integration-test-staging`. Cleans these locations:
 
-* `s3://recover-dev-input-data/staging/`
-* `s3://recover-dev-intermediate-data/staging/json/`
+* `s3://recover-input-data/staging/`
+* `s3://recover-intermediate-data/staging/json/`
 
 
 ## upload-and-deploy-to-prod-main
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -47,13 +47,12 @@ With the current way when the `test_json_to_parquet.py` run, sometimes the glue 
 
 ### integration-test-develop-cleanup
 
-This is responsible for cleaning up any data locations that are used by integration 
+This is responsible for cleaning up any data locations that are used by integration
 tests. This is used after `sceptre-deploy-develop`, but before
 `integration-test-develop`. Cleans these locations:
 
 * `s3://recover-dev-input-data/$GITHUB_REF_NAME/`
 * `s3://recover-dev-intermediate-data/$GITHUB_REF_NAME/json/`
-
 
 
 ### integration-test-develop
@@ -70,7 +69,7 @@ Here that we are **NOT** configuring a S3 event notification configuration for o
 
 ### integration-test-staging-cleanup
 
-This is responsible for cleaning up any data locations that are used by integration 
+This is responsible for cleaning up any data locations that are used by integration
 tests during the staging run. This is used after `sceptre-deploy-staging`, but before
 `integration-test-staging`. Cleans these locations:
 
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -7,12 +7,12 @@ Recover ETL has four github workflows:
 - workflows/codeql-analysis.yml
 - workflows/cleanup.yaml
 
-| Workflow name                     | Scenario it's run |
-| :-------------------------------- |:----------------  |
-| upload-and-deploy                 | on-push from feature branch, feature branch merged into main |
-| upload-and-deploy-to-prod-main    | whenever a new tag is created |
-| codeql-analysis                   | on-push from feature branch, feature branch merged into main
-| cleanup                           | feature branch deleted |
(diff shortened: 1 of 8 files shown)


.github/workflows/README.md
```diff
@@ -57,6 +67,16 @@ This integration test means that you have to wait until the glue workflow has fi
 
 Here that we are **NOT** configuring a S3 event notification configuration for our `prod/staging` space because we plan to submit data to `staging` "manually" after merging a PR into main and triggering the GitHub workflow.
 
+### integration-test-staging-cleanup
+
+This is responsible for cleaning up any data locations that are used by integration
+tests during the staging run. This is used after `sceptre-deploy-staging`, but before
+`integration-test-staging`. Cleans these locations:
+
+* `s3://recover-dev-input-data/staging/`
```
rxu17, line 76: I think this meant to say `recover-input-data` and `recover-intermediate-data`?

src/scripts/manage_artifacts/clean_staging.py
```diff
@@ -0,0 +1,52 @@
...
+    to be kept.
+
+    Arguments:
+        bucket_prefix: The prefix of the bucket to clean. Should end with a '/'
+        bucket: The name of the bucket to clean
+    """
+    print(f"Cleaning bucket: {bucket} with prefix: {bucket_prefix}")
+
+    s3_client = boto3.client('s3')
+    response = s3_client.list_objects_v2(Bucket=bucket, Prefix=bucket_prefix)
```
rxu17, line 26: I'm thinking that we should have an exception handling here for if `bucket_prefix` is `None` or `""` (empty) because I believe `s3_client.list_objects_v2` will list all objects in the bucket if that is the case. Since `parser.add_argument("--bucket_prefix", required=True)` would handle if no `bucket_prefix` is passed perhaps we just need handling for the empty string scenario.

And because we'll be using our prod buckets, we likely want this extra safety net.

rxu17: I wonder if we should also have something here that checks that `main` is not part of the `bucket_prefix` or any of the objects listed and if it is, aborts the process.

rxu17 (approved):

.github/workflows/README.md
BryanFauble: Great catch

src/scripts/manage_artifacts/clean_staging.py
BryanFauble: Thoughts on these 2 guard statements @rxu17 ?
```

    if not args.bucket_prefix or args.bucket_prefix[-1] != '/':
        raise ValueError("Bucket prefix must be provided and end with a '/'")

    if "main" in args.bucket_prefix or "main" in args.bucket:
        raise ValueError(
            "Cannot delete objects in the main directory")
```

philerooski (commented):
"Staging" refers to a specific namespace that's deployed to the prod account, but what you've done (and I think this is a great idea) is to do a clean up step in both dev and prod prior to starting the integration tests. Definitely a nit, but could we rename the script you've written and expand on the docstring to make it clear that this will delete objects in the namespaced location of the intermediate bucket for any namespace, not just staging?

BryanFauble:
> "Staging" refers to a specific namespace that's deployed to the prod account, but what you've done (and I think this is a great idea) is to do a clean up step in both dev and prod prior to starting the integration tests. Definitely a nit, but could we rename the script you've written and expand on the docstring to make it clear that this will delete objects in the namespaced location of the intermediate bucket for any namespace, not just staging?

Absolutely. I will update the name of the script to be `clean_for_integration_test.py` and update the docstring of the script.

src/scripts/manage_artifacts/clean_staging.py
rxu17: Looks good to me! I don't think you need the `"main" in args.bucket` but everything else works. Do you think we should have one more guard statement for the `object_key` and do something like:
```if "main" in object_key:
         raise ValueError(
            "Cannot delete objects in the main directory")
```

BryanFauble: Yeah that makes sense to have that extra:

> guard statement for the object_key later in that function

I will add it in

tests/test_json_to_parquet.py
```diff
@@ -517,6 +541,28 @@ def glue_crawler_role(namespace):
     role_name = f"{namespace}-pytest-crawler-role"
     glue_service_policy_arn = "arn:aws:iam::aws:policy/service-role/AWSGlueServiceRole"
     s3_read_policy_arn = "arn:aws:iam::aws:policy/AmazonS3ReadOnlyAccess"
+
+    # Cleanup if the role/policy already exist
```
rxu17, line 545: Oh, were we running into role/policy duplicates/conflicts here when trying to create role?

BryanFauble: Exactly! This is to make sure we're always running the test from a known state.

PR #122, opened 2024-07-02
[ETL-670] Setup and run pre-commit for all files

BryanFauble:
**Problem:**

1. No linting was set up in the precommit config

**Solution:**

1. Set up the black formatter and isort to run in pre-commit

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,7 @@
+[tool.black]
+line-length = 88
+target-version = ['py39']
+include = '\.pyi?$'
+
+[tool.isort]
+profile = "black"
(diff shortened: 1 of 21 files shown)


pyproject.toml
```diff
@@ -0,0 +1,7 @@
+[tool.black]
+line-length = 88
```
BryanFauble, line 2: 88 is the default, but we can swap to whatever. 88 is what the SynapsePythonClient is using

rxu17 (approved):
Thanks for taking care of this (forgot to include in the compare parquet PR)! LGTM!

PR #123, opened 2024-07-03
[ETL-659] Snowflake git integration

philerooski:
Deploys Snowflake objects to a specific environment as part of our CI/CD process. An environment is analogous to a database. We assume that account-level objects have already been created.

The following objects are created:
* `RECOVER_GIT_REPOSITORY` git repository
* `RECOVER_{environment}` database (I will refer to this merely as the `RECOVER` database from here on for simplicity sake).
* `PARQUET` schema
* `PARQUET_FORMAT` file format.
* `PARQUET_S3` external stage.
* A table for each Parquet dataset in our external stage (manually defined)

Deployment logic and object DDL is organized as a hierarchy:
```
snowflake/objects
└── database
    └── recover
        └── schema
            └── parquet
                ├── file_format
                ├── stage
                └── table
```

(*Almost) every level in this hierarchy has a `deploy.sql` which will deploy all child objects with respect to the current directory.

\* Due to [technical limitations](https://docs.snowflake.com/en/sql-reference/sql/execute-immediate-from#usage-notes), there is no functioning `deploy.sql` file under `snowflake/objects/database`, which would have deployed every database (we only have one database in our deployment, currently).
```
snowflake/objects
├── database
│   ├── deploy.sql
│   └── recover
│       ├── deploy.sql
│       └── schema
│           ├── deploy.sql
│           └── parquet
│               ├── deploy.sql
│               ├── file_format
│               │   ├── deploy.sql
│               │   └── parquet_format.sql
│               ├── stage
│               │   ├── deploy.sql
│               │   └── parquet_s3.sql
│               └── table
│                   ├── deploy.sql
│                   ├── enrolledparticipants_customfields_symptoms_parquet.sql
│                   ├── enrolledparticipants_customfields_treatments_parquet.sql
│                   ├── enrolledparticipants_parquet.sql
│                   ├── fitbitactivitylogs_parquet.sql
│                   ├── fitbitdailydata_parquet.sql
│                   ├── ...
└── deploy.sql

```
For example, the file located at `snowflake/objects/database/recover/deploy.sql` will deploy all objects under the `RECOVER` database, and `snowflake/objects/database/recover/schema/parquet/deploy.sql` will deploy the file formats, stages, and tables under the `PARQUET` schema.

The child objects, that is, the schemas, file formats, stages, tables – anything that is not a database – are defined in such a way as to be agnostic to their parent context. For example, `snowflake/objects/database/recover/schema/parquet/deploy.sql` will deploy the `PARQUET` schema and all its child objects to whichever database your Snowflake user assumes. There is nothing in the SQL which restricts the `PARQUET` schema to be created within the `RECOVER` database. Likewise, the tables can be deployed to any schema, although their DDL is specific to the columns in our Parquet datasets.

The entrypoint for deployment is `snowflake/objects/deploy.sql`. When a commit is pushed to a branch, our CI/CD process will instantiate all Snowflake objects to an environment-specific database prefixed with your branch name, e.g., `RECOVER_ETL_659`. Details on how to manually deploy are provided in the file itself.

## Implementation details
* This PR is an incomplete solution to Snowflake deployment. In a production-ready deployment configuration, we would already have a production database with tables containing data from which we could easily derive tables to deploy to a dev environment. Lacking that, we define our tables from scratch and don't yet load any data into them. Having the ability to create tables from scratch is a critical component for doing reproducible deployments, but it's not an effective production-level deployment strategy. For what work remains to be done to have a fully-featured deployment, see [this comment](https://github.com/Sage-Bionetworks/recover/pull/123#issuecomment-2206756394).
* Objects are created in a few different ways:
  + `CREATE OR REPLACE` will destroy any existing objects and create a new object. This is a powerful, declarative way of enforcing object configurations, but it's not always effective to recreate objects and sometimes existing objects ought to be reused.
  + `CREATE ... IF NOT EXISTS` will create an object if it doesn't yet already exist. We use this statement to preserve the state of that object or its child objects. The prototypical use case for this statement is to preserve tables containing data and their parent objects. I use this statement with `DATABASE` and `SCHEMA` objects because they are parents of table objects. Although, if this is a dev deployment, I instead `CREATE OR REPLACE` the database object in anticipation of cloning preexisting production tables while ensuring a completely reproducible dev environment with each new deployment. 
  + `CREATE OR ALTER` is a mechanism for declaratively updating table objects. More information [here](https://docs.snowflake.com/en/sql-reference/sql/create-table#create-or-alter-table).
* Table data types are defined to be as large as possible without consuming extraneous storage or requiring additional processing. For more information, see [String & binary data types](https://docs.snowflake.com/en/sql-reference/data-types-text) and [Numeric data types](https://docs.snowflake.com/en/sql-reference/data-types-numeric).


snowflake/objects/databases/recover/schemas/parquet/tables/enrolledparticipants_customfields_symptoms_parquet.sql
```diff
@@ -0,0 +1,11 @@
+CREATE OR ALTER TABLE ENROLLEDPARTICIPANTS_CUSTOMFIELDS_SYMPTOMS (
```
thomasyu888, line 1: This is neat, does it automatically put you in the right schema?

thomasyu888 (approved):
🔥 LGTM - with some minor comments - but going to pre-approve!

I wanted to touch on this  "This PR is an incomplete solution to Snowflake deployment. "

I am still a bit confused by this, but could you outline what you think a "production" deployment looks like via CI/CD?  when you push into `main` here, I think it creates a `recover_main` database?

snowflake/objects/databases/recover/schemas/parquet/file_format/parquet_format.sql
```diff
@@ -0,0 +1,7 @@
+/*
+  Create the Parquet file format
+*/
+CREATE OR REPLACE FILE FORMAT {{ parquet_file_format_name }}
```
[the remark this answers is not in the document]
philerooski: It's a tradeoff.

`CREATE OR REPLACE` is the best way to ensure we are deploying in a declarative way without having to worry about state and requires minimal overhead. `FILE FORMAT` is a lightweight object and can be replaced easily.

But with [the caveat](https://docs.snowflake.com/en/sql-reference/sql/create-file-format#usage-notes) about `FILE FORMAT` and `EXTERNAL TABLE`, we either need to enforce top-down that external tables which use `FILE FORMAT` are always recreated during deployment, or – if we were to instead use `CREATE ... IF NOT EXISTS` – we need to depend on the developer to be aware that if they have an existing deployment and they modify the `FILE FORMAT` object, then those changes won't be reflected in subsequent deployments unless they manually delete the object. But if they manually delete the object, then they also have to manually deal with any fallout from the caveat I mentioned above! Alternatively, they could `ALTER` the file format, but already we're getting ourselves into a situation where the developer has to be familiar with and remember this caveat and they have to manage the state of those objects themselves; CI/CD won't do it for them. Of course, we'd have to deal with this complexity when deploying changes to a `FILE FORMAT` in staging or prod, too.

Fortunately, since `EXTERNAL TABLE` objects are merely metadata (schema and table properties), they are also lightweight objects and can be replaced easily. So my preference is to use `CREATE OR REPLACE` for both object types to ensure that we are deploying objects in a state that's reflected in their DDL.

.github/workflows/upload-and-deploy.yaml
```diff
@@ -23,6 +23,59 @@ jobs:
...
+          #mkdir ~/.snowflake
+          #config_file="~/.snowflake/config.toml"
+          config_file=$(mktemp)
+          private_key_file=$(mktemp)
+
+          # Write to the private key file
+          echo "${{ secrets.SNOWFLAKE_PRIVATE_KEY }}" > $private_key_file
+
+          # Write to config.toml file
+          echo 'default_connection_name = "recover"' >> $config_file
```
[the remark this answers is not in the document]
philerooski: Lol. Would be nice if Snowflake had the same mechanism as AWS for authentication.

snowflake/objects/databases/recover/schemas/parquet/tables/enrolledparticipants_customfields_symptoms_parquet.sql
philerooski: If you are deploying from an entrypoint in `database/recover/schema/parquet/deploy.sql` (where you see the `USE SCHEMA` statement) or higher, than yes. But if you are using another schema and you `EXECUTE IMMEDIATE FROM` this file, then this table will be created in that other schema.

philerooski:
> I wanted to touch on this "This PR is an incomplete solution to Snowflake deployment." I am still a bit confused by this, but could you outline what you think a "production" deployment looks like via CI/CD? 

To summarize what the deployment does in this PR, it effectively creates all the objects necessary to create tables (plus a few extra – `FILE FORMAT` and `STAGE` objects – in anticipation of future work) and then creates tables for each of our Parquet datasets. It doesn't load any data, it merely creates the tables (or alters them, if they already exist).

This is less than ideal for a developer. They don't want to load any data and they definitely don't want to do it manually! So in the future, when we have Snowflake tables in production which already contain our data, we can modify the default behavior of this deployment so that we clone those tables, rather than create empty tables. That's what makes this an incomplete solution.

As for what a _production-ready_ deployment system looks like, I believe we ought to have multiple ways of deploying tables:

1. By cloning from existing tables in production
2. By creating the table from scratch and loading data into the table via `COPY INTO` from an external stage.
3. By updating the table if there are changes, and otherwise doing nothing

(1) is our ideal dev environment deployment scenario. It requires minimal overhead and we get a production database in an isolated environment that we can muck around in.
(2) is our bootstrap scenario. If – for whatever reason – we need to redeploy our database entirely from scratch, we should be able to do that.
(3) is our production scenario. We want to update the table metadata if there have been any changes, but data loading happens via a separate mechanism not related to deployment.

> when you push into main here, I think it creates a recover_main database?

No, `RECOVER_MAIN` is one of the databases which we already assume to exist as detailed in `snowflake/objects/deploy.sql`. I use Snowflake scripting to specify that we only create or replace a database if this is not the `main` or `staging` branch/environment.

snowflake/objects/deploy.sql
```diff
@@ -0,0 +1,77 @@
...
+    END IF;
+END;
+$$;
+USE DATABASE IDENTIFIER($database_identifier);
+
+/*
+  Create an external stage over the RECOVER Git repository so that we can
+  use EXECUTE IMMEDIATE FROM statements.
+*/
+CREATE OR REPLACE GIT REPOSITORY recover_git_repository
```
thomasyu888, line 65: I think eventually, we will need ALTER this instead of create or replace in the future.

philerooski: @thomasyu888 Could you explain why we might need to maintain an existing `GIT REPOSITORY` object? By using `CREATE OR REPLACE` we guarantee that our repository is up to date without needing to worry about whether the repository has already been created – which would require a `FETCH` statement.

thomasyu888: @philerooski so, maybe im jumping ahead and just thinking of the `dev` and `prod` deployment.  I guess its not too much lift to create or replace these everytime we push to a branch.

I think I'm struggling a bit conceptually to see the production CI/CD pipeline so I'll wait for you to get the first data flowing in and we can revisit.

snowflake/objects/databases/recover/schemas/parquet/file_format/parquet_format.sql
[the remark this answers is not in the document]
thomasyu888: @philerooski Are we planning on using `EXTERNAL TABLE`?  That's an important caveat you mentioned - but if we aren't, it's probably ok to do this

snowflake/objects/deploy.sql
thomasyu888: Also I guess the issue is there's two parallel tracks here

1. Development workflow
2. Dev and prod deployment

What we are currently tackling here is (2), in that everytime you add a file (lets say Table) or add a scheduled task, the CI/CD is going to run through this entire stack to create the resource without creating all the other resources again. 

For (1), like you've mentioned, it would be CLONE DATABASE, then create all the new resources. 

Is that right?

philerooski: > I guess its not too much lift to create or replace these everytime we push to a branch.

No, it's not much overhead at all. And it's the only way to do stateless/declarative deployment in Snowflake – with the exception of objects which support the CREATE OR ALTER syntax.

> the CI/CD is going to run through this entire stack to create the resource without creating all the other resources again.

I'm a little confused about how you think this works, because you said above that we create or replace most objects every time we push to a branch, but this statement implies that we'll be able to know when a change is independent of already deployed objects, which would require a dependency graph.

> For (1), like you've mentioned, it would be CLONE DATABASE, then create all the new resource per a new branch. We do NOT load all the data up again for existing data., we just load data for the new resources or tables we're adding

That's right. Does it make things clearer if I put it like this: The deployment process for dev/staging/prod is exactly the same _except_ for the hard to replace objects, like tables with data.

snowflake/objects/databases/recover/schemas/parquet/file_format/parquet_format.sql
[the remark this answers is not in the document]
philerooski: @thomasyu888 We are not, but if we were I would still recommend this approach because external tables are easy to replace.

snowflake/objects/deploy.sql
thomasyu888: > we'll be able to know when a change is independent of already deployed objects, which would require a dependency graph.
Hmmm.  lets merge this and see how it plays out - _I think_ we are actually on the same page, but I think need to just _see it_ in action.

Snowflake takes care of what has already been deployed (hence the CREATE... IF NOT EXISTS, or CREATE OR ALTER statements)

philerooski: slight revision: The deployment process for dev/staging/prod is exactly the same except for the hard to replace objects **and their parent objects,** like tables with data, **and their schema and database.**

snowflake/objects/deploy.sql
```diff
@@ -0,0 +1,77 @@
...
+
+    - ROLE `RECOVER_DATA_ENGINEER`
+        * A ROLE with the appropriate privelages to create and deploy
+          objects in a database.
+    - WAREHOUSE `RECOVER_XSMALL`
+    - API INTEGRATION `RECOVER_GIT`
+        * An API INTEGRATION with an API_ALLOWED_PREFIXES value
+          containing 'https://github.com/Sage-Bionetworks/' and
+          API_PROVIDER = GIT_HTTPS_API.
+    - STORAGE INTEGRATION `RECOVER_PROD_S3`
```
BryanFauble, line 16: Do we also want a `recover_dev_s3`? Not sure if we want non-releases to go through any of this.

BryanFauble (approved):
This is awesome. Great work getting this together.

snowflake/objects/deploy.sql
philerooski: We actually do already have a storage integration for our dev account. But it's not a dependency for this deployment, and since in the future we will clone prod tables (or maybe we want to sample them to keep things lightweight?) to their own isolated environment I don't see why we would want to work with the pilot data.

.github/workflows/upload-and-deploy.yaml
```diff
@@ -25,6 +25,59 @@ jobs:
...
+      - name: Configure Snowflake connection
+        run: |
+          # Create temporary files for config.toml and our private key
+          #mkdir ~/.snowflake
+          #config_file="~/.snowflake/config.toml"
+          config_file=$(mktemp)
+          private_key_file=$(mktemp)
+
+          # Write to the private key file
+          echo "${{ secrets.SNOWFLAKE_PRIVATE_KEY }}" > $private_key_file
```
rxu17, line 47: Does `SNOWFLAKE_PRIVATE_KEY` and `SNOWFLAKE_PRIVATE_KEY_PASSPHRASE` ever expire?

snowflake/objects/database/deploy.sql
```diff
@@ -0,0 +1,17 @@
...
+  depth
+  0      deploy.sql =>
+  1        databases/deploy.sql =>
+  2          databases/recover/deploy.sql =>
+  3            databases/recover/schemas/deploy.sql =>
+  4              databases/recover/schemas/parquet/deploy.sql =>
+  5                databases/recover/schemas/parquet/tables/deploy.sql =>
+  6                  databases/recover/schemas/parquet/tables/enrolled_participants.sql
+
+  not possible. To circumvent this issue, we omit the highest level of
```
rxu17, line 14: If I am understanding this correctly to put into more layman terms: 

Instead of executing all database deployments through a single high-level script, we deploy each database individually from the main script (deploy.sql) where from `deploy.sql`, we directly call the lower-level deployment scripts for each database. This approach avoids hitting the maximum depth limit.?

snowflake/objects/database/recover/schema/parquet/table/enrolledparticipants_parquet.sql
```diff
@@ -0,0 +1,92 @@
+CREATE OR ALTER TABLE ENROLLEDPARTICIPANTS (
+    "id" NUMBER(38,0),
+    "CustomFields_Treatments_id" VARCHAR(16777216),
```
rxu17, line 3: Is there a reason we set `VARCHAR` for every str var to be the maximum possible length in Snowflake?

snowflake/objects/deploy.sql
```diff
@@ -0,0 +1,77 @@
...
+  Jira ticket (e.g., etl-123), we substitute a likely hyphen in the `environment`
+  variable passed at runtime.
+*/
+SET safe_environment_identifier = (SELECT REPLACE('&{ environment }', '-', '_'));
+SET database_identifier = CONCAT('recover_', $safe_environment_identifier);
+
+EXECUTE IMMEDIATE
+$$
+BEGIN
+    IF ('&{ environment }' != 'main' AND '&{ environment }' != 'staging') THEN
```
rxu17, line 53: I'm probably missing this somewhere but since we don't have a `staging` branch and I don't see us having an env variable in GH workflow for `staging`, how does it know we are running staging?

rxu17 (approved):
LGTM! I just had some comments and questions

snowflake/README.md
rxu17, line 1: Could we add a section (maybe this should go in confluence docs/contributing readme) on how to contribute to the Snowflake deployment section for the following scenarios:
- Adding a new data type
- Modifying fields in a new data type

Is it as straightforward as just adding a line to `snowflake/objects/database/recover/schema/parquet/table/deploy.sql` and then adding separate script to under: `snowflake/objects/database/recover/schema/parquet/table/` / modifying the respective script for the data type then pushing the changes? Are there other configurations needed?

Do you see other common scenarios where we may need to modify these scripts?

snowflake/objects/database/recover/schema/parquet/table/enrolledparticipants_parquet.sql
philerooski: Table data types are defined to be as large as possible without consuming extraneous storage or requiring additional processing. For more information, see [String & binary data types](https://docs.snowflake.com/en/sql-reference/data-types-text) and [Numeric data types](https://docs.snowflake.com/en/sql-reference/data-types-numeric).

snowflake/objects/database/deploy.sql
philerooski: That's correct.

.github/workflows/upload-and-deploy.yaml
philerooski: They don't expire. Snowflake [supports rotating the key pairs](https://docs.snowflake.com/en/user-guide/key-pair-auth#configuring-key-pair-rotation), but I don't know if we have this configured.

snowflake/objects/deploy.sql
philerooski: Hmm this is an oversight on my part.

Currently, merging into `main` triggers a deployment to our `staging` namespace in AWS. But as it's written now, merging into `main` will trigger a deployment to our `main` environment in Snowflake.

I can think of two solutions off the top of my head:
1. Change the process to match the Snowflake deployment code. This entails doing all PRs into a `staging` branch and then deploying this branch to both `staging` AWS and Snowflake environments.
2. Change the Snowflake deployment code to match the process. Change the logic in our top-level `deploy.sql` so that merging into main does a deployment to the `staging` Snowflake environment as well. Update the `upload-and-deploy-to-prod-main` workflow to do a deployment to the `main` Snowflake environment.

(2) seems significantly simpler to me, so I'd prefer to implement that solution.

rxu17: I agree with (2), and it matches our current deployment workflow for AWS stacks.

snowflake/README.md
philerooski: > Is it as straightforward as

Yep.

It might be a bit premature at this point to start writing docs for our SOPs in Snowflake. It's still an evolving system and this initial PR is primarily meant to set a framework for how deployment is done. At this nascent point in development I think it's best to refer directly to the Snowflake docs if you have any "how do I change this in Snowflake?" questions that aren't related to deployment process.

PR #124, opened 2024-07-06
[ETL-671] Update to use older node version on aws actions

rxu17:
**Purpose:** This is a draft PR because this is just a temp. solution. There is a known discrepancy between the `glibc` version used by the aws glue docker images (2.26) and `glibc` version used by all of the AWS actions  versionswhich is expecting >= 2.27. Until the aws glue docker images are updated to use Node js version 20 (which uses glibc >= 2.27), this is the discovered workaround. [See blog post for more info](https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/). This discussion gives a good view into the issue itself: https://github.com/actions/checkout/issues/1809#issue-2389388431

This solution is still temp. because ideally aws updates their aws glue docker images to use node js 20.

See [JIRA ticket](https://sagebionetworks.jira.com/browse/ETL-671) comments for more detailed updates on what was tried

**Testing:** Our GH actions for aws credentials now works without error

### DIFF
--- a/.github/workflows/upload-and-deploy.yaml
+++ b/.github/workflows/upload-and-deploy.yaml
@@ -14,7 +14,7 @@ env:
   PROD_INPUT_BUCKET: recover-input-data
   PROD_INTERMEDIATE_BUCKET: recover-intermediate-data
   INTEGRATION_TEST_NUM_EXPORTS: 28
-  #ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: true
+  ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: true
 
 jobs:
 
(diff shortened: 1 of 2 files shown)


.github/workflows/upload-and-deploy.yaml
```diff
@@ -14,6 +14,7 @@ env:
   PROD_INPUT_BUCKET: recover-input-data
   PROD_INTERMEDIATE_BUCKET: recover-intermediate-data
   INTEGRATION_TEST_NUM_EXPORTS: 28
+  ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: true
```
thomasyu888, line 17: Hmm. Does it make sense to update to newer versions or is that a harder lift?

rxu17: So the problem is the aws glue docker containers (both glue 4.0 and 3.0) haven't been updated to use node 20 hence they all have node 16 / the glibc version mismatch.

The above would only be a temp. solution though. A more long term solution (until the above gets fixed) i was thinking of and mentioned in the discussion is by building Node.js statically so it doesn't matter what glibc version is used by the docker container. Not sure how viable this is yet so going to try it.

philerooski (approved):

PR #125, opened 2024-07-08
[ETL-635] Add dispatch-to-raw SQS queue

philerooski:


config/develop/namespaced/sqs-dispatch-to-raw.yaml
```diff
@@ -0,0 +1,12 @@
+template:
+  path: sqs-queue.yaml
+parameters:
+  MessageRetentionPeriod: "1209600"
```
thomasyu888, line 4: Odd that these are all strings, but is this an artifact of sceptre?

BryanFauble (approved):

config/develop/namespaced/sqs-dispatch-to-raw.yaml
philerooski: It's [Cloudformation CLI behavior](https://docs.aws.amazon.com/cli/latest/reference/cloudformation/create-stack.html). Parameters must be string, even when they are defined in the template as `Number`, as is the case here.

If I try to use an integer:

```
"Parameter validation failed:\nInvalid type for parameter Parameters[0].ParameterValue, value: 1209600, type: <class 'int'>, valid types: <class 'str'>"
```

rxu17 (approved):
LGTM!

PR #126, opened 2024-07-13
[ETL-670] Adjust cleanup job to run on `main` directory of dev env

BryanFauble:
Problem:

1. We do want to run the cleanup on the `main` directory of the dev branch

Solution:

1. Updating the workflow so the job is triggered and we are able to clean the `main` directory

Testing:

1. Will be looking at the workflow run

### DIFF
--- a/.github/workflows/upload-and-deploy.yaml
+++ b/.github/workflows/upload-and-deploy.yaml
@@ -269,7 +269,6 @@ jobs:
       contents: read
     steps:
       - name: Setup code, pipenv, aws
-        if: github.ref_name != 'main'
         uses: Sage-Bionetworks/action-pipenv-aws-setup@v3
         with:
           role_to_assume: ${{ vars.AWS_CREDENTIALS_IAM_ROLE }}
@@ -281,14 +280,12 @@ jobs:
         run: echo "NAMESPACE=$GITHUB_REF_NAME" >> $GITHUB_ENV
 
       - name: Clean input data bucket
-        if: github.ref_name != 'main'
         run: >
           pipenv run python src/scripts/manage_artifacts/clean_for_integration_test.py
           --bucket $DEV_INPUT_BUCKET
           --bucket_prefix "${{ env.NAMESPACE }}/"
 
       - name: Clean intermediate data bucket
-        if: github.ref_name != 'main'
         run: >
           pipenv run python src/scripts/manage_artifacts/clean_for_integration_test.py
           --bucket $DEV_INTERMEDIATE_BUCKET
--- a/.github/workflows/upload-and-deploy.yaml
+++ b/.github/workflows/upload-and-deploy.yaml
@@ -262,7 +262,6 @@ jobs:
     name: Cleanup non-main branch data before integration tests
     runs-on: ubuntu-latest
     needs: sceptre-deploy-develop
-    if: github.ref_name != 'main'
     environment: develop
     # These permissions are needed to interact with GitHub's OIDC Token endpoint.
     permissions:
@@ -270,6 +269,7 @@ jobs:
       contents: read
     steps:
       - name: Setup code, pipenv, aws
+        if: github.ref_name != 'main'
         uses: Sage-Bionetworks/action-pipenv-aws-setup@v3
         with:
           role_to_assume: ${{ vars.AWS_CREDENTIALS_IAM_ROLE }}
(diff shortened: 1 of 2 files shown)


.github/workflows/upload-and-deploy.yaml
rxu17, line 1: This is actually making me rethink the logic `integration-test-develop`. `integration-test-develop` is just copying the recent 2 weeks of exports and should just be triggering the branch specific namespaced jobs to run on that subset of exports but it shouldn't be doing it for the `main` namespace in the dev project because `main` will always run on all of the data in `dev` to mirror what it does in `prod` (which it's currently doing)

I am thinking now that these are the updates we could make and it makes a bit more sense:
-  `integration-test-develop` needs the `if: github.ref_name != 'main'` line at the highest level
- `integration-test-develop-cleanup` needs the `if: github.ref_name != 'main'` line at the highest level
- `sceptre-deploy-staging` will have `needs: sceptre-deploy-develop`
I think this will give the logic we need. That the integration test and cleanup for when we are in `dev` environment to only run when we are working on the namespaced branch, and doesn't run when we merge to `main`. 

Thoughts?

philerooski (changes requested):
I don't think this is what we want. We don't need to handle integration tests in dev/main any differently than we handle tests for our feature branch. We've already specified `environment` = `develop` on L265, so this is happening separately from our production pipeline and there's no harm in running the integration tests in the main namespace.

BryanFauble:
> We've already specified `environment` = `develop` on L265, so this is happening separately from our production pipeline and there's no harm in running the integration tests in the main namespace.

@philerooski From @rxu17's comments on https://github.com/Sage-Bionetworks/recover/pull/121#discussion_r1663121697 - It sounded like we did not want to run cleanup for `main`.

I agree that we should be running integration tests in the main namespace, but is it acceptable that we would be cleaning up the directories in the main namespace?

philerooski:
@BryanFauble Our dev data has a permanent home in the `pilot-data` prefix in the input bucket. Anything in the `main` prefix is ephemeral and any data in this prefix in any of the buckets ought to reflect HEAD of the `main` branch.

.github/workflows/upload-and-deploy.yaml
philerooski: See the latest comments on this PR, but I think we want the same behavior when there are changes on dev/main as any other dev/* namespace.

rxu17:
> I don't think this is what we want. We don't need to handle integration tests in dev/main any differently than we handle tests for our feature branch. We've already specified environment = develop on L265, so this is happening separately from our production pipeline and there's no harm in running the integration tests in the main namespace.

So the reason I thought we don't need to run integration tests for `main` namespace of `develop` is because we already run it for the `feature_branch` namespace of `develop`. But no nevermind, we do need integration tests for `main` namespace to run so that we are triggering the S3 to JSON job runs after we merge a feature branch for `main` and we don't need to manually run that part. 

In that case, then I think it's fine if we cleanup the `main` namespace of `develop` in both the input and intermediate data buckets. In our case, because all of our pilot data gets copied over (because we only have more than 2 weeks of exports for pilot data), we maintain the expected workflow for `main` namespace which is that the `main` namespaced jobs run on all of the input data

BryanFauble:
Thanks for the feedback @rxu17 @philerooski Take a look at the latest round of changes. We should always clean the directories now in `develop`

src/scripts/manage_artifacts/clean_for_integration_test.py
```diff
@@ -56,9 +54,6 @@ def main() -> None:
     if not args.bucket_prefix or args.bucket_prefix[-1] != "/":
         raise ValueError("Bucket prefix must be provided and end with a '/'")
 
-    if "main" in args.bucket_prefix:
```
rxu17, line 59: Ah I forgot about us having to remove this if we want to clean up `main`. Hmm, should we add another safeguard?
```
if `dev` not in `args.bucket`:
   raise ValueError("Must be using buckets in the develop environment")
```

OR

```
dev_buckets = [`recover-dev-input-data`, 'recover-dev-intermediate-data`, 'recover-dev-processed-data', 'recover-dev-cloudformation']

if `args.bucket` not in dev_buckets:
   raise ValueError("Must be using buckets in the develop environment")
```
Overkill?

rxu17 (approved):
LGTM! Just one comment!

src/scripts/manage_artifacts/clean_for_integration_test.py
BryanFauble: We will also be running this on the prod buckets too - But the `staging` directory of those buckets. I think controlling this in the workflow makes the most sense here.

philerooski: We could add a safeguard for prod/main (we still want to be able to clean up prod/staging). I think this script should be agnostic to the object key, except for the aforementioned case.

BryanFauble: How does this guard condition look?

https://github.com/Sage-Bionetworks/recover/pull/126/commits/6e366260fdaa97dffded87acc1ecadde13787c1b

rxu17: Sorry, my brain is not functioning right now. Yes I second what Phil said.

src/scripts/manage_artifacts/clean_for_integration_test.py
```diff
@@ -56,8 +54,10 @@ def main() -> None:
...
-        raise ValueError("Cannot delete objects in the main directory")
+    if "main" in args.bucket_prefix and "recover-dev" not in args.bucket:
```
rxu17, line 57: 🔥

philerooski (approved):

PR #127, opened 2024-07-15
[ETL-659] Enable any branch to be deployed to any environment

philerooski:
When merging a PR into main, we deploy changes to a `staging` Snowflake environment. Because we can only do variable substitution in a stage path using CLI variables when invoking our entrypoint `snowflake/objects/deploy.sql` from the command line, I've split the `environment` variable into an `environment` and `git_branch` variable. Both must be specified.


BryanFauble (approved):
Great changes! It's awesome seeing this all come together.

thomasyu888 (approved):
🔥 LGTM!

PR #128, opened 2024-07-16
[ETL-670] Update code to paginate over all results

BryanFauble:
**Problem:**

1. There is a maximum of 1000 objects that can return from a single call to `list_objects_v2`

**Solution:**

1. Using a pagination to loop over all objects in the bucket/prefix

**Testing:**

1. To review the action runs


philerooski (approved):
This works 👍 You might have also invoked the CLI directly 
```
aws s3 rm s3://<bucket-name>/<path> --exclude <exclusion-pattern> --recursive
```

PR #130, opened 2024-07-17
Fix Snowflake cleanup

philerooski:
I did not use the right reference

### DIFF
--- a/.github/workflows/cleanup.yaml
+++ b/.github/workflows/cleanup.yaml
@@ -6,7 +6,6 @@ jobs:
 
   snowflake-delete:
     name: Delete Snowflake development environment
-    if: ${{ github.ref_name != 'main' && github.ref_name != 'staging' }}
     runs-on: ubuntu-latest
     env:
       PRIVATE_KEY_PASSPHRASE: ${{ secrets.SNOWFLAKE_PRIVATE_KEY_PASSPHRASE }}
@@ -52,7 +51,7 @@ jobs:
         run: |
           snow sql \
             -q "
-              SET safe_environment_identifier = (SELECT REPLACE('$GITHUB_REF_NAME', '-', '_'));
+              SET safe_environment_identifier = (SELECT REPLACE('${{ github.event.ref }}', '-', '_'));
               SET database_identifier = CONCAT('recover_', \$safe_environment_identifier);
               DROP DATABASE IDENTIFIER(\$database_identifier);
             "
--- a/.github/workflows/cleanup.yaml
+++ b/.github/workflows/cleanup.yaml
@@ -6,6 +6,7 @@ jobs:
 
   snowflake-delete:
     name: Delete Snowflake development environment
+    if: ${{ github.event.ref != 'main' && github.event.ref != 'staging' }}
     runs-on: ubuntu-latest
     env:
       PRIVATE_KEY_PASSPHRASE: ${{ secrets.SNOWFLAKE_PRIVATE_KEY_PASSPHRASE }}


.github/workflows/cleanup.yaml
```diff
@@ -52,7 +51,7 @@ jobs:
         run: |
           snow sql \
             -q "
-              SET safe_environment_identifier = (SELECT REPLACE('$GITHUB_REF_NAME', '-', '_'));
+              SET safe_environment_identifier = (SELECT REPLACE('${{ github.event.ref }}', '-', '_'));
```
thomasyu888, line 54: I haven't looked very closely, but we will want to be careful to not bring down main / dev databases and tables. Only for feature branches in the future.

.github/workflows/cleanup.yaml
```diff
@@ -6,7 +6,6 @@ jobs:
 
   snowflake-delete:
     name: Delete Snowflake development environment
-    if: ${{ github.ref_name != 'main' && github.ref_name != 'staging' }}
```
rxu17, line 9: Why was this removed?

.github/workflows/cleanup.yaml
philerooski: That could only happen if we delete the `main` branch, which would be catastrophic in other ways. But no need to compound the damage. I added a conditional on this job just in case.

.github/workflows/cleanup.yaml
philerooski: We want to use `github.event.ref` instead.

rxu17: sorry i meant the entire conditional but i see it back now

.github/workflows/cleanup.yaml
```diff
@@ -52,7 +52,7 @@ jobs:
         run: |
           snow sql \
             -q "
-              SET safe_environment_identifier = (SELECT REPLACE('$GITHUB_REF_NAME', '-', '_'));
+              SET safe_environment_identifier = (SELECT REPLACE('${{ github.event.ref }}', '-', '_'));
```
rxu17, line 55: I'm looking at examples of `github.event.ref` and it seems to indicate that it includes the full reference path not just the branch name e.g `refs/heads/main`?

philerooski: @rxu17 The naming convention would make more sense if it did that, but you can see its actual behavior in this line where we use `github.event.ref` to substitute the namespace/branch name: https://github.com/Sage-Bionetworks/recover/actions/runs/9963030161/job/27528123692#step:3:7

rxu17: Ah i didn't notice this subtle difference before. Does the behavior of `github.event.ref` change if it's used in the conditional context or it just spits out the exact same thing as `github.ref_name` in this case as well?

```
snowflake-delete:
    name: Delete Snowflake development environment
    if: ${{ github.event.ref != 'main' && github.event.ref != 'staging' }}
```

Because usually we use `if: github.ref_name != 'main'`

philerooski: > it just spits out the exact same thing as github.ref_name

This appears to be the case. There is no such thing as `github.event.ref_name`. See the available properties here: https://docs.github.com/en/rest/using-the-rest-api/github-event-types?apiVersion=2022-11-28#event-object-common-properties

rxu17: Right, `github.ref_name` exists here: https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

rxu17 (approved):
LGTM!

PR #131, opened 2024-07-17
Specify default warehouse in Snowflake config

philerooski:
So we don't run into silly errors.


rxu17 (approved):

philerooski:
Build is failing because the feature branch name is too long. I'm going to merge since I did not touch sceptre/AWS at all in this PR.

PR #134, opened 2024-07-25
[ETL-674] Add script to compress intermediate JSON

philerooski:
Recombines the NDJSON part sets which we wrote to our intermediate bucket and uploads a gzipped NDJSON to an adjacent location.

## Data Input
Each part set is located at:

  `s3://<bucket-name>/<namespace>/json/dataset={dataset}/cohort={cohort}/`

where `{dataset}` is the dataset identifier (datatype),  and `{cohort}` is either 'adults_v1' or 'pediatric_v1'. Each file in the part set is named:

  `{file_identifier}.part{part_number}.ndjson`

## Data Output
Each combined and gzipped part set is written to:

  `s3://<bucket-name>/<namespace>/compressed_json/dataset={dataset}/cohort={cohort}/{file_identifier}.ndjson.gz`

### DIFF
--- a/src/scripts/compress_json/compress_json.py
+++ b/src/scripts/compress_json/compress_json.py
@@ -1,0 +1,202 @@
+"""
+This script combines and compresses NDJSON part files in our intermediate S3 bucket
+into a single gzipped NDJSON file per part set.
+
+Which S3 bucket is used is dependent on whether `--testing` is passed.
+
+-- Data Input --
+Each part set is located at:
+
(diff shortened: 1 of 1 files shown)


philerooski:
@BryanFauble I amended my commit but it seems like SonarCloud ran its analysis on the previous version of the code again.

BryanFauble:
> @BryanFauble I amended my commit but it seems like SonarCloud ran its analysis on the previous version of the code again.

I am not sure if SonarCloud would work properly with amended commits. I suspect that is needs a new commit hash for sonarcloud to run.

PR #135, opened 2024-08-20
[ETL-673] Upload compressed JSON as part of S3 to JSON

philerooski:
In addition to partitioning and writing JSON to JSON datasets as "part files", e.g., 
```
FitbitIntradayCombined_20220111-20230103.part0.ndjson
FitbitIntradayCombined_20220111-20230103.part1.ndjson
...
```
Write the JSON as a gzip archive to a separate `compressed_json` S3 prefix.
```
FitbitIntradayCombined_20220111-20230103.ndjson.gz
```

This PR is meant to ease the transition towards exclusively using compressed JSON data as the input to JSON to Parquet. Hence, there are a few changes here which are meant to be refactored later on once we remove the logic for writing uncompressed part files. The primary changes are to the function `write_file_to_json_dataset` with some supporting changes in helper functions.

You can get a more concrete feel for how this affects the behavior of S3 to JSON by examining the result of the integration tests in `s3://recover-dev-intermediate-data/etl-673/`.

### DIFF
--- a/src/glue/jobs/s3_to_json.py
+++ b/src/glue/jobs/s3_to_json.py
@@ -526,13 +526,14 @@ def write_file_to_json_dataset(
     Returns:
         list: A list of files uploaded to S3
     """
-    # Configuration related to where we write our part files
+    # Configuration related to where we write the JSON
     output_dir = os.path.join(
         f"dataset={metadata['type']}", f"cohort={metadata['cohort']}"
     )
     os.makedirs(output_dir, exist_ok=True)
     part_number = 0
-    part_output_path = get_output_path(
+    # we update this value as we write
+    current_part_path = get_output_path(
         metadata=metadata, part_number=part_number, output_dir=output_dir, touch=True
     )
     compressed_output_path = get_output_path(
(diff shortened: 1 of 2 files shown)


BryanFauble (approved):
LGTM!

philerooski:
There was a bug where I was uploading the files before closing their file objects, but I believe that's resolved now.

src/glue/jobs/s3_to_json.py
```diff
@@ -522,74 +526,91 @@ def write_file_to_json_dataset(
...
 
     # We will attach file metadata to the uploaded S3 object
     s3_metadata = _derive_str_metadata(metadata=metadata)
+    line_count = 0
     uploaded_files = []
-    with z.open(json_path, "r") as input_json:
-        current_output_path = output_path
-        line_count = 0
+    # fmt: off
+    # <python3.10 requires this backslash syntax, we currently run 3.7
```
thomasyu888, line 548: Thanks, we do have a ticket to update to python 3.10, do you think we should do that sooner than later?

thomasyu888 (approved):
🔥 LGTM! Thanks for the review and work here!

src/glue/jobs/s3_to_json.py
philerooski: No, we may end up abandoning Glue altogether so it doesn't make sense spending time upgrading python just to avoid these minor inconveniences.

PR #136, opened 2024-08-21
bug squashed: uploading part file before file stream closed

philerooski:
I did not catch every instance in my previous PR where I was uploading files before the corresponding stream was closed. I also added an additional logging statement to verify that we are deleting files after upload.


thomasyu888 (approved):
🔥 LGTM! Thanks for catching this!

BryanFauble (approved):

PR #137, opened 2024-08-22
[ETL-675] Update Glue tables to use compressed JSON

philerooski:


BryanFauble (approved):
LGTM!

templates/glue-tables.j2
```diff
@@ -51,17 +51,16 @@ Resources:
...
         PartitionKeys: {{ dataset.partition_keys }}
         StorageDescriptor:
           Columns: {{ dataset.columns }}
-          Compressed: false
+          Compressed: true
           InputFormat: org.apache.hadoop.mapred.TextInputFormat
           Location: !Sub s3://${S3IntermediateBucketName}/${Namespace}/${S3IntermediateBucketPrefix}/dataset={{ dataset["type"] }}
           OutputFormat: org.apache.hadoop.hive.ql.io.HiveIgnoreKeyTextOutputFormat
-          StoredAsSubDirectories: false
+          StoredAsSubDirectories: true
```
rxu17, line 63: Curious, why did this get updated to `true` and wasn't true before? I assume that since everything in technically in a subdirectory

rxu17 (approved):
LGTM! Just a question

templates/glue-tables.j2
philerooski: Ah, yeah should have mentioned this in the description. Glue doesn't use this parameter, so it doesn't really matter what we set it to, but if we were to use this table with Athena we would want to specify the correct parameter value `true`, since we do use subdirectories. I thought updating the value made more sense than deleting it outright.

PR #138, opened 2024-08-30
[ETL-636] Raw Lambda

philerooski:
Adds resources:
- Lambda: The raw Lambda, which polls the dispatch-to-raw SQS queue and writes compressed JSON to the raw S3 bucket.
- IAM Role: An IAM role for the above Lambda.

Raw Lambda compresses JSON data contained in an export (zip archive) from the input S3 bucket and uploads it to the raw S3 bucket. It makes heavy use of Python file objects and multipart uploads and can download/compress/upload with a relatively low, fixed memory overhead with respect to the size of the uncompressed JSON. For example, during testing I was able to process a 450 MB (uncompressed) JSON file with a max memory used of 345 MB.

### DIFF
--- a/config/develop/namespaced/sqs-dispatch-to-raw.yaml
+++ b/config/develop/namespaced/sqs-dispatch-to-raw.yaml
@@ -3,7 +3,7 @@ template:
 parameters:
   MessageRetentionPeriod: "1209600"
   ReceiveMessageWaitTimeSeconds: "20"
-  VisibilityTimeout: "120"
+  VisibilityTimeout: "900"
   SNSTopicSubscription: !stack_output_external "{{ stack_group_config.namespace }}-sns-dispatch::SnsTopicArn"
 dependencies:
   - develop/namespaced/sns-dispatch.yaml
--- a/src/lambda_function/raw/app.py
+++ b/src/lambda_function/raw/app.py
@@ -1,7 +1,26 @@
+"""
+Raw Lambda
+
+A module for an AWS Lambda function that compresses JSON data contained in an
+export (zip archive) from S3 and uploads it to the raw S3 bucket. This module
+makes heavy use of Python file objects and multipart uploads and can
+decompress/compress/upload with a relatively low, fixed memory overhead
+with respect to the size of the uncompressed JSON.
+
+Example Usage:
+The module is intended to be deployed as an AWS Lambda function that listens to
+events from the `dispatch-to-raw` SQS queue.
+
+Environment Variables:
+- `RAW_S3_BUCKET`: The raw S3 bucket name where compressed data will be stored.
+- `RAW_S3_KEY_PREFIX`: The S3 key prefix within the raw bucket where data is written.
+"""
+
 import gzip
 import io
 import json
 import logging
+import os
 import zipfile
 
 import boto3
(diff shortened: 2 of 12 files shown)


config/develop/namespaced/sqs-dispatch-to-raw.yaml
```diff
@@ -3,7 +3,7 @@ template:
 parameters:
   MessageRetentionPeriod: "1209600"
   ReceiveMessageWaitTimeSeconds: "20"
-  VisibilityTimeout: "120"
+  VisibilityTimeout: "900"
```
thomasyu888, line 6: Can you speak to the increase here?

src/lambda_function/raw/app.py
```diff
@@ -0,0 +1,337 @@
...
+    Returns:
+        str: An S3 key in the format:
+            `{raw_key_prefix}/dataset={data_type}/cohort={cohort}/{basename}.ndjson.gz`
+
+    Notes:
+        - The function expects the input `key` to be formatted as `{namespace}/{cohort}/{export_basename}`.
+        - The data type is derived from the first underscore-delimited component of the file basename.
+    """
+    key_components = key.split("/")
+    # input bucket keys are formatted like `{namespace}/{cohort}/{export_basename}`
```
thomasyu888, line 70: Just checking, but is it always formatted like this?

src/lambda_function/raw/app.py
```diff
@@ -0,0 +1,337 @@
...
+        #     {
+        #       'ETag': 'etag-value',
+        #       'PartNumber': 2
+        #     },
+        #   'upload_id': 'existing-upload-id',
+        #   'part_number': 2
+        # }
+
+    Notes:
+        - Parts must be larger than AWS minimum requirements (typically 5 MB),
```
thomasyu888, line 149: Did you need to configure anything to make sure the parts are > 5MB?

src/lambda_function/raw/app.py
```diff
@@ -0,0 +1,337 @@
...
+                    if compressed_data.tell() >= part_threshold:
+                        yield compressed_data_wrapper(
+                            compressed_data=compressed_data, chunk_number=chunk_number
+                        )
+                        compressed_data.seek(0)
+                        compressed_data.truncate(0)
+                        chunk_number = chunk_number + 1
+            yield compressed_data_wrapper(
+                compressed_data=compressed_data, chunk_number=chunk_number
+            )
```
thomasyu888, line 242: I'm a bit confused here, is this double recursion?

thomasyu888 (approved):
🔥 LGTM! Going to pre-approve, but not sure if @rxu17 had any last comments?

config/develop/namespaced/sqs-dispatch-to-raw.yaml
philerooski: We don't know yet how long it could take for this Lambda to run on prod-sized data. Especially given the relatively small memory I've allotted. So I've set the Lambda timeout to it's maximum (900). Since the Lambda could take up to 900 seconds to complete, we also need the associated SQS event to show up in the queue again even if the Lambda fails at 899 seconds, hence we set the visibility timeout on the SQS queue to the timeout of the Lambda. AWS actually enforces this server side, so the deployment will fail if we use a smaller value.

src/lambda_function/raw/app.py
philerooski: @thomasyu888 It could change in the future, but that would also mean changing every other stack that tries to read data from the input bucket, so no harm in hard coding this format.

src/lambda_function/raw/app.py
philerooski: That's taken care of by `yield_compressed_data`. The `part_threshold` is set to 8 MB by default.

src/lambda_function/raw/app.py
thomasyu888: Lets hope it doesn't change in the future.

src/lambda_function/raw/app.py
```diff
@@ -0,0 +1,337 @@
...
+            raw_key_prefix=raw_key_prefix,
+        )
+        # Stream the zip file from S3
+        with io.BytesIO() as object_stream:
+            s3_client.download_fileobj(
+                Bucket=sns_message["Bucket"],
+                Key=sns_message["Key"],
+                Fileobj=object_stream,
+            )
+            object_stream.seek(0)
```
BryanFauble, line 306: This is super cool how you can do this

BryanFauble (approved):
LGTM!

PR #139, opened 2024-09-06
[ETL-616] Implement Great Expectations to run on parquet data

rxu17:
### **Purpose:** 
This draft PR adds the Great Expectations (GX) Parquet Glue jobs to the Recover ETL workflow. When JSON to Parquet workflow finishes, this will run the GX job per data type.

Currently this just supports running GX with expectations for the `fitbitdailydata` and `healthkitv2workouts` datasets. All other data types will have their jobs error out.

### **Changes:**
Highlights big changes

**New Code:**
- [run_great_expectations_on_parquet.py](https://github.com/Sage-Bionetworks/recover/blob/etl-616/src/glue/jobs/run_great_expectations_on_parquet.py) : This is the script to be run by the GX on Parquet jobs. There are some workarounds in the script notably in the `add_data_docs_sites` and `add_validation_results_to_store` functions to allow us to add validation results to the validation store and also have the data docs (our GX report) render them since we use a [EphemeralDataContext](https://docs.greatexpectations.io/docs/reference/api/data_context/ephemeraldatacontext_class/) context object without having to create checkpoints, a GX config file (which would likely have us have to confirm to a specific GX repo structure), etc. If we prefer to switch to using a more persistent data content object like [FileDataContext](https://docs.greatexpectations.io/docs/reference/api/data_context/FileDataContext_class), that could be explored further in this ticket.
-  [data values validation suite](https://github.com/Sage-Bionetworks/recover/blob/etl-616/src/glue/resources/data_values_expectations.json): this will be where we manually add our expectations to in the future. I find this way the easiest way because we have to look through the expectations gallery in order to find the expectation we want to add and in case we want to be able to validate our outputted set of expectations against this list.

**Changes to old Code:**
- [Updated glue workflow triggers](https://github.com/Sage-Bionetworks/recover/blob/etl-616/templates/glue-workflow.j2#L281) : because there is a limitation to the number of jobs per trigger (50 max), I split the previous trigger into two, one for each set of jobs for CompareParquet and the new GX on Parquet jobs.
- [Pinned dependency urllib3<2 as part of additional python modules](https://stackoverflow.com/questions/76414514/cannot-import-name-default-ciphers-from-urllib3-util-ssl-on-aws-lambda-us): There is a compatibility issue with `boto3` and `urllib3 2.0`. We need to pin `urllib3<2`

### **Tests:**

- [x] Integration testing in AWS (currently running)
- [x] Unit tests
- [X] Tests that script can produce reports from [sample validation suite](https://github.com/Sage-Bionetworks/recover/blob/etl-616/src/glue/resources/data_values_expectations.json) in the shareable artifacts bucket

**Viewable reports at (with AWS VPN turned on):**
- [fitbitdailydata gx report](https://recover-dev-shareable-artifacts-vpn.s3.amazonaws.com/etl-616/great_expectation_reports/fitbitdailydata/parquet/index.html)
- [healthkitv2workouts gx report](https://recover-dev-shareable-artifacts-vpn.s3.amazonaws.com/etl-616/great_expectation_reports/healthkitv2workouts/parquet/index.html)

**Sample screenshots of a report:**
image

image

EDIT: Also added tests to our CI/CD as previously they weren't automatically running all of them each time


thomasyu888:
Great work here - took a quick glance, and will review more tomorrow.  I'll leave it to Phil for a final review.

src/glue/resources/data_values_expectations.json
thomasyu888, line 1: Just curious, how did you generate this?

thomasyu888 (commented):
🔥 LGTM!  I'm going to pre-approve, but I'll leave it to @philerooski to do a final review.

src/glue/jobs/run_great_expectations_on_parquet.py
thomasyu888, line 1: So when this is deployed, will the docs be served and updated daily on `.../main/...html`

src/glue/resources/data_values_expectations.json
rxu17: By hand, i just added from the google sheet of some expectations that was provided to the expected structure that great expectations wanted for expectations.

src/glue/jobs/run_great_expectations_on_parquet.py
rxu17: Yes, that is the simple plan for now

templates/glue-workflow.j2
```diff
@@ -287,7 +295,36 @@ Resources:
...
+          {% for dataset in datasets if "Garmin" in dataset["data_type"] %}
+          - JobName: !Sub "${Namespace}-{{ dataset["stackname_prefix"] }}-Job"
+            State: SUCCEEDED
+            LogicalOperator: EQUALS
+          {% endfor %}
+        Logical: AND
+      StartOnCreation: true
+      WorkflowName: !Ref JsonToParquetWorkflow
+
+  GreatExpectationsParquetTrigger:
```
philerooski, line 311: We should make the trigger data type specific. That way if one of the JSON to Parquet jobs fails we still run the GX job for those that succeeded. We can use Jinja formatting to create each trigger for us.

rxu17: So essentially we'd have something for the GX jobs that mirrors what is currently going on with the JSON to Parquet jobs in this workflow template then because they are all linked to run one after another?

src/glue/jobs/run_great_expectations_on_parquet.py
```diff
@@ -0,0 +1,408 @@
...
+
+    # Load the JSON file with the expectations
+    logger.info("reads_expectations_from_json")
+    expectations_data = read_json(
+        s3=s3,
+        s3_bucket=args["cfn_bucket"],
+        key_prefix=args["expectation_suite_key_prefix"],
+    )
+    logger.info("adds_expectations_from_json")
+    add_expectations_from_json(
```
philerooski, line 379: In an ideal world, rather than logging the error instead of raising it, only deploy the GX job and associated trigger for a data type if that data type has expectations. That would require us to do some Jinja on `src/glue/resources/data_values_expectations.json` (sort of like how we do Jinja on `src/glue/resources/table_columns.json` in `templates/glue-tables.j2`). But if that's too complicated then we can log the error and return early.

templates/glue-workflow.j2
philerooski: Yeah, sort of. It will look a little funny because each JSON to Parquet job will have a trigger watching it, and that trigger will kick off the respective GX job. That's like 10x the number of triggers we have in our current JSON to Parquet workflow. Since both GX jobs and JSON to Parquet jobs use the processed data bucket for their temporary files, we'll want to watch out for exceeding S3 quotas (which can happen if too many writes are happening on a single bucket).

src/glue/jobs/run_great_expectations_on_parquet.py
rxu17: Yup it looks very doable. I'll add this and [that](https://github.com/Sage-Bionetworks/recover/pull/139#discussion_r1747794348) in.

templates/glue-workflow.j2
```diff
@@ -300,6 +308,34 @@ Resources:
...
+      Name: !Sub "${Namespace}-{{ dataset['stackname_prefix'] }}GreatExpectationsParquetTrigger"
+      Actions:
+        - JobName: !Sub ${Namespace}-{{ dataset["stackname_prefix"] }}-GreatExpectationsParquetJob
+          Arguments:
+            "--data-type": {{ dataset["data_type"].lower() }}
+            "--namespace": !Ref Namespace
+            "--cfn-bucket": !Ref CloudformationBucketName
+            "--parquet-bucket": !Ref ParquetBucketName
+            "--shareable-artifacts-bucket": !Ref ShareableArtifactsBucketName
+            "--expectation-suite-key-prefix": !Sub "${Namespace}/src/glue/resources/data_values_expectations.json"
```
philerooski, line 325: This is the key, not the key prefix, no?

templates/glue-workflow.j2
```diff
@@ -300,6 +308,34 @@ Resources:
...
+            "--additional-python-modules": "great_expectations~=0.18,urllib3<2"
+      Description:  This trigger runs the great expectation parquet job for this data type after completion of the JSON to Parquet job for this data type
+      Type: CONDITIONAL
+      Predicate:
+        Conditions:
+          - JobName: !Sub "${Namespace}-{{ dataset['stackname_prefix'] }}-Job"
+            State: SUCCEEDED
+            LogicalOperator: EQUALS
+        Logical: AND
```
philerooski, line 334: This is optional if there is only one condition https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-glue-trigger-predicate.html

config/develop/glue-job-role.yaml
```diff
@@ -6,5 +6,6 @@ parameters:
   S3IntermediateBucketName: {{ stack_group_config.intermediate_bucket_name }}
   S3ParquetBucketName: {{ stack_group_config.processed_data_bucket_name }}
   S3ArtifactBucketName: {{ stack_group_config.template_bucket_name }}
+  S3ShareableArtifactBucketName: {{ stack_group_config.shareable_artifacts_vpn_bucket_name }}
```
philerooski, line 9: Do we need to update this line in the corresponding file in `config/prod`?

config/develop/namespaced/glue-job-run-great-expectations-on-parquet.yaml
```diff
@@ -0,0 +1,18 @@
...
+  - develop/glue-job-role.yaml
+stack_name: "{{ stack_group_config.namespace }}-glue-job-RunGreatExpectationsParquet"
+parameters:
+  Namespace: {{ stack_group_config.namespace }}
+  JobDescription: Runs great expectations on a set of data
+  JobRole: !stack_output_external glue-job-role::RoleArn
+  TempS3Bucket: {{ stack_group_config.processed_data_bucket_name }}
+  S3ScriptBucket: {{ stack_group_config.template_bucket_name }}
+  S3ScriptKey: '{{ stack_group_config.namespace }}/src/glue/jobs/run_great_expectations_on_parquet.py'
+  GlueVersion: "{{ stack_group_config.json_to_parquet_glue_version }}"
```
philerooski, line 13: It's not necessarily the case that JSON to Parquet and this GX job will always use the same Glue version. My recommendation is to hardcode the version here and in `config/prod` since it's unlikely to be referenced elsewhere, but you could add an additional variable to the config file if you prefer.

config/develop/namespaced/glue-workflow.yaml
```diff
@@ -19,7 +20,10 @@ parameters:
   CompareParquetMainNamespace: "main"
   S3SourceBucketName: {{ stack_group_config.input_bucket_name }}
   CloudformationBucketName: {{ stack_group_config.template_bucket_name }}
+  ShareableArtifactsBucketName: {{ stack_group_config.shareable_artifacts_vpn_bucket_name }}
+  ExpectationSuiteKeyPrefix: "{{ stack_group_config.namespace }}/src/glue/resources/data_values_expectations.json"
```
philerooski, line 24: same comment, this is a key, not a key prefix. Also update in `config/prod`.

.github/workflows/upload-and-deploy.yaml
```diff
@@ -133,10 +133,12 @@ jobs:
         run: |
           pipenv install ecs_logging~=2.0
 
-      - name: Test lambda scripts with pytest
+      - name: Test scripts with pytest (lambda, etc.)
         run: |
           pipenv run python -m pytest tests/test_s3_event_config_lambda.py -v
           pipenv run python -m pytest tests/test_s3_to_glue_lambda.py -v
+          pipenv run python -m pytest tests/test_lambda_dispatch.py -v
+          pipenv run python -m pytest tests/test_consume_logs.py -v
```
philerooski, line 141: Should these each be their own separate steps?

.github/workflows/upload-and-deploy.yaml
```diff
@@ -257,7 +259,8 @@ jobs:
         if: matrix.tag_name == 'aws_glue_4'
         run: >
           su - glue_user --command "cd $GITHUB_WORKSPACE &&
-          python3 -m pytest tests/test_json_to_parquet.py --namespace $NAMESPACE -v"
+          python3 -m pytest tests/test_json_to_parquet.py --namespace $NAMESPACE -v &&
+          python3 -m pytest tests/test_run_great_expectations_on_parquet.py -v"
```
philerooski, line 263: ditto separate steps

src/glue/jobs/run_great_expectations_on_parquet.py
```diff
@@ -0,0 +1,408 @@
...
+
+    Returns:
+        pyspark.sql.dataframe.DataFrame: spark dataframe of the read in parquet dataset
+    """
+    s3_parquet_path = f"s3://{parquet_bucket}/{namespace}/parquet/dataset_{data_type}/"
+    dynamic_frame = glue_context.create_dynamic_frame_from_options(
+        connection_type="s3",
+        connection_options={"paths": [s3_parquet_path]},
+        format="parquet",
+    )
```
philerooski, line 209: Any reason you didn't `create_dynamic_frame_from_catalog`?

philerooski (approved):
There is a change needed in `config/prod/glue-job-role.yaml`, but everything else is minor.

.github/workflows/upload-and-deploy.yaml
rxu17: Actually they can all be grouped under one command this way they are run all at once, and if one set of tests have failures, the other tests in another test suite still get run.

.github/workflows/upload-and-deploy.yaml
rxu17: See comment above. Also particularly for this one, since we're using aws glue 4.0 for `test_run_great_expectations_on_parquet.py` for now, I will keep this under here.

src/glue/jobs/run_great_expectations_on_parquet.py
rxu17: No particular reason. I see now that the `create_dynamic_frame.from_catalog` will have the benefit of not dealing with schema inferencing. Were there any pitfalls you see with using `create_dynamic_frame.from_options` over `create_dynamic_frame.from_catalog`?

config/develop/namespaced/glue-job-run-great-expectations-on-parquet.yaml
rxu17: Do you envision other GX jobs to use all the same Glue version (if they use glue)?

src/glue/jobs/run_great_expectations_on_parquet.py
philerooski: I tend to use each command in different scenarios (`from_catalog` in JSON to Parquet and `from_options` in pyspark when I don't have a Glue table over the data already). Now that I've typed that, I immediately see that you want to use `from_options` since we don't have Glue tables over our Parquet.

config/develop/namespaced/glue-job-run-great-expectations-on-parquet.yaml
philerooski: Yes.

PR #140, opened 2024-09-10
[ETL-683] Increase memory of raw lambda

philerooski:
Changes:
* Increased memory from 1024 to 1769 (1 vCPU).
* Put `compressed_data` in a context manager so that we don't have to manually close
* Removed return statement at end of `main` in case we ever want to use a batch size larger than 1.

Details about benchmarking exploration are [here](https://sagebionetworks.jira.com/browse/ETL-683?focusedCommentId=219741).

### DIFF
--- a/src/lambda_function/raw/template.yaml
+++ b/src/lambda_function/raw/template.yaml
@@ -52,7 +52,7 @@ Resources:
       Handler: app.lambda_handler
       Runtime: !Sub "python${LambdaPythonVersion}"
       Role: !Ref RoleArn
-      MemorySize: 1024
+      MemorySize: 1769
       EphemeralStorage:
         Size: 2048
       Timeout: 900
--- a/src/lambda_function/raw/app.py
+++ b/src/lambda_function/raw/app.py
@@ -211,35 +211,37 @@ def yield_compressed_data(object_stream: io.BytesIO, path: str, part_threshold=N
         part_threshold = 8 * 1024 * 1024
     with zipfile.ZipFile(object_stream, "r") as zip_stream:
         with zip_stream.open(path, "r") as json_stream:
-            compressed_data = io.BytesIO()
-            # analogous to the part number of a multipart upload
-            chunk_number = 1
-            with gzip.GzipFile(
-                filename=os.path.basename(path),
-                fileobj=compressed_data,
-                compresslevel=6,
-                mode="wb",
-            ) as gzip_file:
-                # We can expect at least 10x compression, so reading/writing the
-                # JSON in 10*part_threshold chunks ensures we do not flush the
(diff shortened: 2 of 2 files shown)


thomasyu888 (commented):
Great benchmarking!  I'll leave this to @rxu17 for a review.  

👀 the nesting of context managers

.github/workflows/upload-and-deploy.yaml
```diff
@@ -132,11 +132,13 @@ jobs:
       - name: Install additional python dependency
         run: |
           pipenv install ecs_logging~=2.0
+          pipenv install pytest-datadir
 
       - name: Test lambda scripts with pytest
         run: |
           pipenv run python -m pytest tests/test_s3_event_config_lambda.py -v
           pipenv run python -m pytest tests/test_s3_to_glue_lambda.py -v
+          pipenv run python -m pytest -v tests/test_lambda_raw.py
```
rxu17, line 141: Nit: Depending on if this gets merged first or mine (may be a slight merge conflict), but i think you can link the tests like:

```
- name: Test scripts with pytest (lambda, etc.)
        run: |
          pipenv run python -m pytest \
            tests/test_s3_event_config_lambda.py \
            tests/test_s3_to_glue_lambda.py \
            tests/test_lambda_dispatch.py \
            tests/test_consume_logs.py \
            tests/test_lambda_raw.py -v
```

src/lambda_function/raw/template.yaml
```diff
@@ -52,7 +52,7 @@ Resources:
       Handler: app.lambda_handler
       Runtime: !Sub "python${LambdaPythonVersion}"
       Role: !Ref RoleArn
-      MemorySize: 1024
+      MemorySize: 1769
```
rxu17, line 55: Nit: Something i was thinking of but do we want to make this a variable instead? So the develop version of the lambda can declare a different memory size than the prod version? Or were you running into this for both?

rxu17 (approved):
LGTM! Just a few minor comments

.github/workflows/upload-and-deploy.yaml
philerooski: Eh, it's better for the commands to be discrete and explicit, imo. I'm not sure if the behavior is the same running the tests one at a time (separate `pipenv run...`) vs running all the tests at one time (a single `pipenv run...`). But if there are advantages (other than brevity) I'm open to running all the tests at one time.

src/lambda_function/raw/template.yaml
philerooski: I don't think so. This is not an expensive function to run, so no harm in keeping the configurations the same.

PR #141, opened 2024-09-17
[ETL-611] Raw sync lambda

philerooski:
# Raw Sync Lambda

This Lambda verifies that the input and raw S3 buckets are synchronized. It's triggered by a Cloudwatch Events rule at midnight UTC each day.

This is accomplished by verifying that all non-zero sized JSON in each export in the input S3 bucket have a corresponding object in the raw S3 bucket. Because we only download the central directory, located near the end of a zip archive, verification can be done extremely quickly and without needing to download most of the export. If a JSON file from an export is found to not have a corresponding object in the raw bucket, the export is submitted to the raw Lambda (via the dispatch SNS topic) for processing.

### DIFF
--- a/src/lambda_function/raw_sync/README.md
+++ b/src/lambda_function/raw_sync/README.md
@@ -1,0 +1,38 @@
+# Raw Sync Lambda
+
+The raw sync Lambda ensures that the input and raw S3 buckets are synchronized
+by verifying that all non-zero sized JSON in each export in the input bucket
+have a corresponding object in the raw bucket.
+
+If a JSON file from an export is found to not have a corresponding object in the raw bucket,
+the file is submitted to the raw Lambda (via the dispatch SNS topic) for processing.
+
+## Development
+
+The Serverless Application Model Command Line Interface (SAM CLI) is an
+extension of the AWS CLI that adds functionality for building and testing
+Lambda applications.
+
+To use the SAM CLI, you need the following tools.
+
+* SAM CLI - [Install the SAM CLI](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-sam-cli-install.html)
+* Docker - [Install Docker community edition](https://hub.docker.com/search/?type=edition&offering=community)
+
+You may need the following for local testing.
+* [Python 3 installed](https://www.python.org/downloads/)
+
+You will also need to configure your AWS credentials, if you have not already done so.
+
+## Creating a local build
+
+Use the SAM CLI to build and test your lambda locally.
+Build your application with the `sam build` command.
+
+```bash
+cd src/lambda_function/raw_sync/
+sam build
+```
+
+## Tests
+
+Tests are available in `tests/test_lambda_raw_sync.py`.
--- a/src/lambda_function/raw_sync/app.py
+++ b/src/lambda_function/raw_sync/app.py
@@ -1,0 +1,436 @@
+"""
+Raw Sync Lambda
+
+This script verifies that the input and raw S3 buckets are synchronized.
+
+This is accomplished by verifying that all non-zero sized JSON in each export
+in the input S3 bucket, excepting "Manifest.json", have a corresponding object
+in the raw S3 bucket.  Because we only download the central directory, typically
+located near the end of a zip archive, verification can be done extremely quickly
+and without needing to download most of the export.
+
+If a JSON file from an export is found to not have a corresponding object in the raw bucket,
+the export is submitted to the raw Lambda (via the dispatch SNS topic) for processing.
+"""
+
+import logging
+import os
+import struct
+import zipfile
+from collections import defaultdict
+from io import BytesIO
+from typing import Optional
+
+import boto3
+
+logger = logging.getLogger()
+logger.setLevel(logging.INFO)
+
+
+def lambda_handler(event: dict, context: dict) -> None:
+    """
+    Entrypoint for this Lambda.
+
+    Args:
+        event (dict)
+        context (dict): Information about the runtime environment and
+            the current invocation
+    """
+    s3_client = boto3.client("s3")
+    input_bucket = os.environ.get("INPUT_S3_BUCKET")
+    input_key_prefix = os.environ.get("INPUT_S3_KEY_PREFIX")
+    raw_bucket = os.environ.get("RAW_S3_BUCKET")
+    raw_key_prefix = os.environ.get("RAW_S3_KEY_PREFIX")
+    main(
+        event=event,
+        s3_client=s3_client,
+        input_bucket=input_bucket,
+        input_key_prefix=input_key_prefix,
+        raw_bucket=raw_bucket,
+        raw_key_prefix=raw_key_prefix,
+    )
+
+
+def list_s3_objects(
+    s3_client: boto3.client, bucket: str, key_prefix: str, key_format: str
+) -> dict:
+    """
+    Recursively list all objects under an S3 bucket and key prefix which
+    conform to a specified format.
+
+    It's assumed that all objects under `key_prefix` have a key prefix
+    themselves which conforms to one of two formats:
+
+    "input" format: `{namespace}/{cohort}/
+    "raw" format: `{namespace}/json/dataset={data_type}/cohort={cohort}/`
+
+    Args:
+        s3_client (boto3.client): An S3 client
+        bucket (str): The name of the S3 bucket.
+        key_prefix (str): The S3 key prefix to recursively list files from.
+        key_format (str): The format used by the keys, either "input" or "raw"
+
+    Returns (dict): A dictionary where each hierarchy corresponds to the
+        ordering of the variables in the `key_format`, excepting {namespace}.
+
+        For example:
+
+        `key_format`="raw":
+
+            {
+                "data_type_one": {
+                    "cohort_one": [
+                        "object_one",
+                        "object_two",
+                        ...
+                    ],
+                    "cohort_two": [
+                        "object_one",
+                        ...
+                    ]
+                }
+                "data_type_two": {
+                    ...
+                },
+                ...
+            }
+
+        `key_format`="input":
+
+            {
+                "cohort_one": [
+                    "object_one",
+                    "object_two",
+                    ...
+                ],
+                "cohort_two": [
+                    "object_one",
+                    ...
+                ]
+            }
+    """
+    paginator = s3_client.get_paginator("list_objects_v2")
+    response_iterator = paginator.paginate(Bucket=bucket, Prefix=key_prefix)
+    if key_format == "raw":
+        result = defaultdict(lambda: defaultdict(list))
+    elif key_format == "input":
+        result = defaultdict(list)
+
+    for response in response_iterator:
+        for obj in response.get("Contents", []):
+            key = obj["Key"]
+            key_components = key.split("/")
+            if key_format == "raw":
+                try:
+                    data_type = next(
+                        part.split("=")[1]
+                        for part in key_components
+                        if part.startswith("dataset=")
+                    )
+                    cohort = next(
+                        part.split("=")[1]
+                        for part in key_components
+                        if part.startswith("cohort=")
+                    )
+                    result[data_type][cohort].append(key)
+                except StopIteration:
+                    # Skip keys that don't match the expected pattern
+                    continue
+            elif key_format == "input" and len(key_components) == 3:
+                cohort = key_components[1]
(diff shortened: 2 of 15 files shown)


src/lambda_function/raw_sync/app.py
```diff
@@ -0,0 +1,518 @@
...
+        key (str): The S3 object key to process.
+        key_format (str): The format of the key, either "raw" or "input".
+        result (dict): The dictionary where keys are appended. For the "raw" format, it is a
+                       nested dictionary structured as result[data_type][cohort]. For "input",
+                       it is structured as result[cohort].
+
+    Returns:
+        None
+    """
+    if not key.endswith("/"):  # Ignore keys that represent "folders"
```
thomasyu888, line 73: Nit: Is there a better way to achieve this?

src/lambda_function/raw_sync/app.py
```diff
@@ -0,0 +1,518 @@
...
+                    for part in key_components
+                    if part.startswith("cohort=")
+                )
+                result[data_type][cohort].append(key)
+            except StopIteration:
+                # Skip keys that don't match the expected pattern
+                return
+        elif key_format == "input" and len(key_components) == 3:
+            cohort = key_components[1]
+            result[cohort].append(key)
```
thomasyu888, line 93: Does this function not return anything?  Are you taking advantage of the mutable dict that you pass in through memory?  I feel like that adds vulnerability in the code

src/lambda_function/raw_sync/app.py
```diff
@@ -0,0 +1,518 @@
...
+    paginator = s3_client.get_paginator("list_objects_v2")
+    response_iterator = paginator.paginate(Bucket=bucket, Prefix=key_prefix)
+    if key_format == "raw":
+        result = defaultdict(lambda: defaultdict(list))
+    elif key_format == "input":
+        result = defaultdict(list)
+    for response in response_iterator:
+        for obj in response.get("Contents", []):
+            key = obj["Key"]
+            append_s3_key(
```
thomasyu888, line 163: It may be better to return the values to append here

thomasyu888 (approved):
🔥 LGTM, I'm going to let @rxu17 do a review here, but I took a look at the tests and they look good.

The most concern I have is around data quality and making sure that this change doesn't introduce bugs in the production data.  I wonder if we need GX truly running in production so we can be certain changes don't add unintended bugs.

src/lambda_function/raw_sync/app.py
```diff
@@ -0,0 +1,518 @@
...
+
+    Args:
+        body (bytes): The byte content from which to extract EOCD fields.
+        eocd_offset (int): The offset position of the EOCD signature in the body.
+
+    Returns:
+        tuple: A tuple containing (central_directory_offset, central_directory_size).
+            Both are int type.
+    """
+    eocd_fields = struct.unpack("<4s4H2LH", body[eocd_offset : eocd_offset + 22])
```
rxu17, line 263: Is it possible to get an invalid `eocd_offset` (not -1 but something else)? Do we have handling for that?

src/lambda_function/raw_sync/app.py
```diff
@@ -0,0 +1,518 @@
...
+        if adjusted_range_size > total_size * 2:
+            logger.error(
+                "Did not find an end of central directory record in "
+                f"s3://{bucket}/{key}"
+            )
+            return []
+        logger.warning(
+            f"Calling this function recursively with `range_size` = {adjusted_range_size}"
+        )
+        return list_files_in_archive(
```
rxu17, line 380: Given the current recursion limit in python being 1000, in the scenario that we hit it, are we concerned about `range_size` being multiplied 1000 times/large?

tests/test_lambda_raw_sync.py
rxu17, line 1: A few tests that I think would be good to have (I might have missed some tests so please let me know):

- A test that checks to see if StopIteration exception is met in `append_s3_key` function and the dict of s3 keys is not changed
- Two more tests for `match_corresponding_raw_object` function that tests when you'd get a None return so there's comprehensive coverage:
  - When cohort not found
  - When data_type not found
- A test for `list_files_in_archive` function when we end up with empty list of files

src/lambda_function/raw_sync/app.py
```diff
@@ -0,0 +1,518 @@
...
+    raw_keys = list_s3_objects(
+        s3_client=s3_client,
+        bucket=raw_bucket,
+        key_prefix=raw_key_prefix,
+        key_format="raw",
+    )
+    for export_key in sum(export_keys.values(), []):
+        # input bucket keys are formatted like `{namespace}/{cohort}/{export_basename}`
+        namespace, cohort = export_key.split("/")[:2]
+        file_list = list_files_in_archive(
```
rxu17, line 483: technically, should this ever be []? Should we be throwing some warning?

rxu17 (commented):
Just did a first pass and had some comments/questions. Great work so far!

src/lambda_function/raw_sync/README.md
```diff
@@ -0,0 +1,38 @@
+# Raw Sync Lambda
+
+The raw sync Lambda ensures that the input and raw S3 buckets are synchronized
+by verifying that all non-zero sized JSON in each export in the input bucket
+have a corresponding object in the raw bucket.
+
+If a JSON file from an export is found to not have a corresponding object in the raw bucket,
+the file is submitted to the raw Lambda (via the dispatch SNS topic) for processing.
```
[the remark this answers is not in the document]
philerooski: That's a possibility. Assuming the raw Lambda is working correctly, that could mean a few things:

1. The raw Lambda was throttled, or it failed for normal resource constraint reasons, in which case we do want to re-submit. 
2. The file is unable to be processed by the raw Lambda in its current configuration. Probably this means we've encountered an extremely large JSON file, probably ~120-150 GB uncompressed, or 3 times the size of anything we've received before. If we don't handle this case, the consequences are that we will submit this behemoth for processing once a day where it will timeout after 15 minutes. Alternatively, perhaps a non-zip file or some other corrupted file is accidentally deposited in the input bucket, which would be ignored by the raw-sync Lambda after it determines it's not a zip-file (although this process would also repeat each day until the corrupted file is removed). We don't want to re-submit the file, but I can't think of a graceful way for the raw-sync lambda to differentiate between this case and case (1).

In short... I can't think of a way for the raw-sync Lambda to gracefully handle files which simply can't be processed by raw Lambda. It could help to have raw-sync Lambda publish to an SNS topic which sends us an email when it encounters a file which is out-of-sync, at which point we could manually intervene to see what happened. In the interim we could watch the raw Lambda's DLQ to see if there are any files which repeatedly fail to be processed.

src/lambda_function/raw_sync/app.py
philerooski: Not that I know of.

src/lambda_function/raw_sync/app.py
philerooski: The function modifies the `result` dict in-memory.

> I feel like that adds vulnerability in the code

How so? I could make a copy of the dict, update the copy, and then return -- but what's the benefit?

src/lambda_function/raw_sync/app.py
philerooski: Not according to [the docs](https://docs.python.org/3/library/stdtypes.html#bytes.rfind). See `determine_eocd_offset` for where that method is used.

src/lambda_function/raw_sync/app.py
philerooski: That would probably mean a very large file (which is almost certainly not a zip file) was dropped in the input bucket. How large? 32*2^1000 KB, or 3.5e290 petabytes, much larger than any object in S3 could ever be.

So we will never reach the recursion limit and would instead timeout after 15 minutes of trying to download enough of the object to find the EOCD marker.

src/lambda_function/raw_sync/app.py
philerooski: CE shouldn't be sending us empty zip files, but agnostic of what's going on in the broader context of the pipeline, a function like `list_files_in_archive` should return an empty list if there are no files in the archive.

If we want a mechanism to alert us of suspiciously empty exports or other data validation issues with data in the input bucket, I don't think the raw-sync Lambda is the right context for that.

tests/test_lambda_raw_sync.py
```diff
@@ -0,0 +1,533 @@
...
+            "namespace/cohort_one/object_two",
+        ],
+        "cohort_two": ["namespace/cohort_two/object_one"],
+    }
+
+    result = app.list_s3_objects(s3_client, bucket_name, key_prefix, key_format)
+    assert result == expected_output, f"Expected {expected_output}, but got {result}"
+
+
+def test_match_corresponding_raw_object(mocked_raw_keys):
```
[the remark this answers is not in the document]
philerooski: Sure, although I'm not sure what you mean by run the tests simultaneously.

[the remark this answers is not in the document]
rxu17: Oh I just meant when pytest is called, the tests will be run independent of each other. Here it's dependent on the previous test passing before the following condition can be run.

src/lambda_function/raw_sync/app.py
rxu17: Could there be a scenario where `eocd_offset` was determined incorrectly (-1) for a non-large file?

src/lambda_function/raw_sync/app.py
rxu17: I guess I'm also thinking about if somehow the logic in `list_files_in_archive` was inaccurate after updating it or what not, leading to a [] file. Should we flag that in `list_files_in_archive`?

src/lambda_function/raw_sync/README.md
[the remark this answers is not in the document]
rxu17: I can see how the 2nd scenario would be difficult to catch. I don't remember the retry limit but yeah if it ends up in the DLQ, that can be a good area to check in the interim.

src/lambda_function/raw_sync/app.py
philerooski: It can't be a zip file without the EOCD byte marker, so it's not possible to have a false negative. Is that what you mean?

src/lambda_function/raw_sync/app.py
philerooski: Are you thinking of what if there is a false negative? We return an empty list in two cases:

1. The EOCD marker is not found and the next `range_size` value we want to attempt is greater than 2x the object size, meaning we just downloaded the entire object and didn't find an EOCD marker. Hence, this is not a valid zip file and there's no file list to return.
2. We find an EOCD marker (this is a valid zip file), but either the zip file is empty, or it contains files which we don't want to process anyway according to our conditions:
```
"/" not in zip_info.filename  # necessary for pilot data only
and "Manifest" not in zip_info.filename
and zip_info.file_size > 0
```

We do `logger.error` in the case of (1), because we shouldn't pass non-zip files to `list_files_in_archive`. It could be useful to do something like `logger.warning` if `file_list` is still empty after (2), because an empty zip-file or a zip-file containing no useful data is pretty unusual, although it doesn't prevent `list_files_in_archive` from doing exactly what we expect of it. If there is a bug in the logic that the tests didn't catch, than a `logger.warning` could help us detect that.

src/lambda_function/raw_sync/app.py
philerooski: Actually, I did think of a small benefit. Returning a copy makes the variable update in `list_s3_objects` more explicit, and creating a shallow copy within the function can be done without any additional imports and with just one extra line of code.

src/lambda_function/raw_sync/app.py
rxu17: Yup

src/lambda_function/raw_sync/app.py
rxu17: Sounds good to me!

rxu17 (approved):
LGTM!

src/lambda_function/raw_sync/app.py
thomasyu888: This is probably more relevant when making mutable default arguments, but I prefer the way you've written it to make the function self contained
https://docs.python-guide.org/writing/gotchas/#:~:text=Python's%20default%20arguments%20are%20evaluated,to%20the%20function%20as%20well.

PR #143, opened 2024-09-19
[ETL-688] Handle "Deleted" data types correctly in raw lambda

philerooski:
The raw Lambda did not differentiate between “Deleted” and non-deleted datatypes. For example, HealthKitV2Heartbeat_Samples_Deleted_20230701-20230702.ndjson.gz was written to the HealthKitV2Heartbeat S3 prefix. I fixed that.


thomasyu888 (approved):
🔥 LGTM!

PR #144, opened 2024-10-02
[ETL-693] Add function for parsing data type to raw-sync lambda

philerooski:
### Raw-Sync Lambda

I pulled the logic in `main` which derives the data type from the file name into its own function so that we can write tests for it. I also did some linting.


rxu17 (approved):
LGTM!

PR #146, opened 2024-10-03
Use raw key prefix to construct raw key

philerooski:
Fixes a bug introduced in #144

### DIFF
--- a/src/lambda_function/raw_sync/app.py
+++ b/src/lambda_function/raw_sync/app.py
@@ -483,14 +483,16 @@ def get_data_type_from_path(path: str) -> str:
     return data_type
 
 
-def get_expected_raw_key(namespace: str, data_type: str, cohort: str, path: str) -> str:
+def get_expected_raw_key(
+    raw_key_prefix: str, data_type: str, cohort: str, path: str
+) -> str:
     """Get the expected raw S3 key
 
     Get the expected raw S3 key of a raw bucket object corresponding to the given
     input bucket object.
 
     Args:
-        namespace (str): The namespace of the corresponding input object.
+        raw_key_prefix (str): The namespaced S3 prefix where raw objects are written.
         data_type (str): The data type of the corresponding input object.
         cohort (str): The cohort of the corresponding input object.
         path (str): The path of the file relative to the zip archive (export).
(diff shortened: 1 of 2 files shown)


tests/test_lambda_raw_sync.py
thomasyu888, line 1: Thanks for adding this fix, but may I ask why this wasn't caught initially?  Are we missing integration tests?

rxu17: This technically was caught during validation after the staging integration ran as part of the original ticket etl-693 stemmed from: https://sagebionetworks.jira.com/browse/ETL-611

thomasyu888: Thanks @rxu17, glad to see that the validation steps are helpful. I guess I'm wondering if there's something that could've prevented this before a merge and run on staging, OR are we relying on multiple testing and validation processes to ensure validity in the pipeline?

src/lambda_function/raw_sync/app.py
```diff
@@ -500,7 +502,7 @@ def get_expected_raw_key(namespace: str, data_type: str, cohort: str, path: str)
     """
     file_identifier = os.path.basename(path).split(".")[0]
     expected_key = (
-        f"{namespace}/json/dataset={data_type}"
+        f"{raw_key_prefix}/dataset={data_type}"
```
rxu17, line 505: I'm still confused on the issue and how switching to `raw_key_prefix` fixed it. Isn't `raw_key_prefix` just `S3RawKeyPrefix: "{{ stack_group_config.namespace }}/json/"` (after tracing back to the stack configs)? It's the same as `{namespace}/json` in the end

tests/test_lambda_raw_sync.py
philerooski: @thomasyu888 this bug only appeared in the case where we are syncing different namespaces, which is not a real use case, but is something which happens incidentally as part of [the validation instructions](https://sagebionetworks.jira.com/browse/ETL-611?focusedCommentId=220789) where we check that the entire prod dataset can be processed in one go.

src/lambda_function/raw_sync/app.py
philerooski: > It's the same as {namespace}/json in the end

Yes, but only in the case where the namespace of the input bucket is the same as the namespace of the raw bucket. That's always going to be true in the typical use case of raw-sync Lambda, but since I implemented it in such a way that we can sync different namespaces, we need to make sure that we are referencing the appropriate namespace for the raw bucket when constructing the expected raw key.

rxu17 (approved):
LGTM!

PR #147, opened 2024-10-04
Log formatting fixes for raw-sync lambda

philerooski:
Some of the paths in our log messages contained double forward slashes in the path because they were not being formatted correctly. Any path constructions have been replaced with `os.path.join(...)` calls.

### DIFF
--- a/src/lambda_function/raw_sync/app.py
+++ b/src/lambda_function/raw_sync/app.py
@@ -375,7 +375,7 @@ def list_files_in_archive(
         if adjusted_range_size > total_size * 2:
             logger.error(
                 "Did not find an end of central directory record in "
-                f"s3://{bucket}/{key}"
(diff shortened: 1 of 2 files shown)


src/lambda_function/raw_sync/app.py
rxu17, line 1: Do you need to add/modify a test case where you have more than one slash

rxu17 (approved):
LGTM just one comment!

src/lambda_function/raw_sync/app.py
philerooski: @rxu17 I added a test case for `get_expected_raw_key`. I generally don't test logging output since it's usually not important to the correctness of the code, although I do typically test for at least the existence of `error` or other critical logs that could be important in determining a silent failure type situation. For example, not finding the end of central directory record because a non-zip file was deposited in the input bucket.

PR #148, opened 2024-10-14
[ETL-694] Specify python-version in actions/setup-python

philerooski:
Fixes bug described in ticket.

### DIFF
--- a/.github/workflows/upload-and-deploy-to-prod-main.yaml
+++ b/.github/workflows/upload-and-deploy-to-prod-main.yaml
@@ -16,7 +16,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: actions/setup-python@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: ${{ env.PYTHON_VERSION }}
       - uses: pre-commit/action@v3.0.0
 
 
@@ -125,7 +127,7 @@ jobs:
           cat $SNOWFLAKE_CONFIG_PATH
 
       - name: Install Snowflake CLI
-        uses: Snowflake-Labs/snowflake-cli-action@v1
+        uses: Snowflake-Labs/snowflake-cli-action@v1.5
         with:
           default-config-file-path: ${{ env.SNOWFLAKE_CONFIG_PATH }}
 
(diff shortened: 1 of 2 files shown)


BryanFauble (approved):

philerooski:
There appears to be another bug with the Snowflake action we are using, as described here: https://github.com/Snowflake-Labs/snowflake-cli-action/issues/28

I think this will be resolved by Snowflake-Labs within the next few days so let's wait on them, rather than specifying a specific version of the action, which would need to be reverted later.

thomasyu888:
> There appears to be another bug with the Snowflake action we are using, as described here: https://github.com/Snowflake-Labs/snowflake-cli-action/issues/28
> 
> I think this will be resolved by Snowflake-Labs within the next few days so let's wait on them, rather than specifying a specific version of the action, which would need to be reverted later.

Lets wait a week to see if it gets resolved by snowflake labs.  If not we don't want it to block our work - will the GX work be able to be resumed?

philerooski:
@thomasyu888 

> will the GX work be able to be resumed?

Technically this is not a blocker, since we aren't making any changes to Snowflake we don't need to worry about breaking anything Snowflake related. We can merge regardless.

But if we wait a few days to see how this plays out, we can see all the green check marks before merging without impeding any dev work, since the GX work is not related to Snowflake, either.

philerooski:
Snowflake-Labs used an unfortunate tag `v1` for their initial tag, rather than `v1.0`. This caused `@v1` to point to a specific tag, rather than use the latest minor version of 1.x. The fix at this point seems to be to specify a specific tag, `@v1.5`, unless Snowflake-Labs fixes the `v1` tag, but it's not clear when they might do so.

thomasyu888:
> Snowflake-Labs used an unfortunate tag `v1` for their initial tag, rather than `v1.0`. This caused `@v1` to point to a specific tag, rather than use the latest minor version of 1.x. The fix at this point seems to be to specify a specific tag, `@v1.5`, unless Snowflake-Labs fixes the `v1` tag, but it's not clear when they might do so.

Make sense - unfortunate... Agree that lets move forward with V1.5 tag

philerooski:
@thomasyu888 Can I get an approving review? Sorry for the formality.

PR #150, opened 2024-10-16
Enable raw sync events rule

philerooski:
#141 Did not have the events rule which triggers the raw sync lambda enabled before merging. This PR rectifies that.


thomasyu888 (approved):
🤖 LGTM!

PR #153, opened 2024-11-14
[ETL-690] Add additional expectations for fitbit data types

philerooski:
There is an issue with the upload-and-deploy workflow that is causing tests to fail, which I've documented here: https://sagebionetworks.jira.com/browse/ETL-699

I was able to deploy this manually and did not have any issues generating validation docs for the additional data types. Additionally, I ran the tests which were prevented from running due to the issue mentioned above and they all passed (as expected -- there weren't any changes made to the code).

### DIFF
--- a/src/glue/resources/data_values_expectations.json
+++ b/src/glue/resources/data_values_expectations.json
@@ -1,6 +1,18 @@
 {
+    "fitbitactivitylogs": {
+        "expectations": [
+            {
+                "expectation_type": "expect_column_values_to_be_between",
+                "kwargs": {
+                    "column": "AverageHeartRate",
+                    "min_value": 25,
+                    "max_value": 300
+                }
+            }
+        ],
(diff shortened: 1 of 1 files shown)


rxu17 (approved):
Pre-approving! 

I noticed that `upload-and-deploy` is not included in the build but it [ran with an error here](https://github.com/Sage-Bionetworks/recover/actions/runs/11844333197), that's different than error [presented here](https://sagebionetworks.jira.com/browse/ETL-699).

thomasyu888 (approved):
🔥 LGTM!

philerooski:
@rxu17 

> I noticed that upload-and-deploy is not included in the build checks anymore

That's strange. I can see upload-and-deploy. 

The workflow completes successfully now.

rxu17:
> That's strange. I can see upload-and-deploy.

Yup, it appears now for me - after your recent deployments/pushes

